PR #4, opened 2022-05-17
Support for DeletePattern (pruner)

jsmorph:
The basic approach: remove deleted patterns from match results and rebuild the matcher periodically. More specifically:

1. Remember patterns that have been added
2. Remember patterns that have been removed (implicitly)
3. Filter `MatchesFor...()` results to remove any removed patterns
4. Rebuilding the matcher state periodically with only the live patterns
5. Maintain some statistics to help decide when to rebuild

See the directory `pruner`, which has a README.


pruner/README.md
```diff
@@ -0,0 +1,27 @@
+# `DelPattern`
+
+The core quamina.Matcher doesn't currently support deleting patterns.
+Some of the contemplated implementations of would probably be pretty
```
timbray, line 4: missing words after "of"?

pruner/README.md
```diff
@@ -0,0 +1,27 @@
...
+difficult.  At least one approach is pretty easy: Wrap the current
+matcher to filter removed patterns from match results and periodically
+rebuild the matcher from scrach with the live patterns.  More
+specifically:
+
+1. Remember patterns that have been added
+2. Remember patterns that have been removed (implicitly)
```
timbray, line 11: what does "implicitly" mean… I think you have to remember them so you can filter them out of Match results?

pruner/doc.go
```diff
@@ -0,0 +1,15 @@
+// Package pruner supports removing a pattern from a matcher.
+//
+// The core quamina.Matcher doesn't currently support deleting
+// patterns.  Some of the contemplated implementations of would
```
timbray, line 4: missing words after "of"?

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+
+import (
+	"sync"
+	"time"
+
+	quamina "quamina/lib"
+)
+
+// Stats reports basic counts to aid in deciding when to Rebuild.
+type Stats struct {
```
timbray, line 11: A bit troubling that the word "pattern" doesn't appear in the name either of the type nor the fields.  When I'm reading code it seems like it would be nice to see things like `if stats.patternsEmitted` or `if patternStats.emitted` or some such.  I have been accused of being overly fond of long names…

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+	// Some of these values are ints instead of uints because Go
+	// likes to print uints in hex, and I'd like to see some
+	// temporary logging output in decimal.  ToDo: back to uints?
```
timbray, line 14: I'd stay with ints.  Probably means you have to write an is-positive check in your API

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+//
+// Eventually automatically-invoked rebuild policies might be
+// pluggable.
+type Matcher struct {
+	// Matcher is the underlying matcher that does the hard work.
+	//
+	// Maybe Matcher should maybe not be embedded or public.
+	*quamina.Matcher
+
+	// live is live set of patterns.
```
timbray, line 66: once again, might be nice if these were called patternState & patternStats

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+//
+//   MinAction is less than the sum of counts of found and filtered
+//   patterns and
+//
+//   FilteredToEmitted is greater than the ratio of counts of filtered
+//   and found patterns.
+//
+// defaultRebuildTrigger provides the default trigger policy used by
+// NewMatcher.
+type tooMuchFiltering struct {
```
timbray, line 97: why *64?  I like to constrain the compiler as little as possible unless the semantics get in the way… I'd probably just say `float` and `int`.  Seems unlikely to me that anyone would want to require more than 2**32 matches…

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func newTooMuchFiltering(ratio float64, min int64) *tooMuchFiltering {
+	return &tooMuchFiltering{
+		FilteredToEmitted: ratio,
+		MinAction:         min,
+	}
+}
+
+func (t *tooMuchFiltering) Rebuild(added bool, s *Stats) bool {
```
timbray, line 109: what does "added" mean here?  Comment might be helpful?

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+		// filtered patterns -- I think.
+		return false
+	}
+
+	if s.Emitted+s.Filtered < t.MinAction {
+		return false
+	}
+
+	if s.Emitted == 0 {
+		return true
```
timbray, line 122: hm? Having trouble sorting this out which means a comment might be nice?

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+	if m.rebuildTrigger.Rebuild(added, &m.stats) {
+		return m.rebuild(added)
+	}
+
+	return nil
+}
+
+// AddPattern calls the underlying AddPattern method and then maybe
+// rebuilds the index (if the AddPattern succeeded).
+func (m *Matcher) AddPattern(x quamina.X, pat string) error {
```
timbray, line 190: could we waste 4 bytes and say "pattern"?

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+	// Remove any X that isn't in the live set.
+
+	acc := make([]quamina.X, 0, len(xs))
+
+	// We're updating stats.Filtered, so we need a write lock.  If
+	// we forget about stats.Filtered, we can live with only a
+	// read lock here.
+	var emitted, filtered int64
+	for _, x := range xs {
+		have, err := m.live.Get(x)
```
timbray, line 242: could the variable names be a bit more descriptive?  I think "have" means something like "filterThisMatch" or "patternIsDeleted"

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if have == "" {
```
timbray, line 246: So, I can't say `addPattern("", {"a": [1]}` ?

Maybe use a `*string` or add an `ok` return value to signal the value is not there?

pruner/pruner.go
```diff
@@ -0,0 +1,339 @@
...
+
+	// Nothing fancy here now.
+
+	var (
+		then = time.Now()
+		m1   = quamina.NewMatcher()
+	)
+
+	if fearlessly {
+		// Let the GC reduce heap requirements?
```
timbray, line 305: I see, the idea is not to have two Matchers in memory. On the other hand, we worry about a failed rebuild() destroying a working Matcher.   We'll know which is the right thing to worry about after some more experience...

pruner/state.go
```diff
@@ -0,0 +1,81 @@
+package pruner
+
+import (
+	"fmt"
+	"sync"
+
+	quamina "quamina/lib"
+)
+
+// State represents the required capabilities for maintaining the set
```
timbray, line 10: Insert grumble about short name here…

pruner/state.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Get returns the pattern for the given X.
+	//
+	// Since a pattern can't be the empty string, that zero value
+	// indicates no corresponding pattern.
+	Get(x quamina.X) (string, error)
+}
+
+// MemState is a State that is just a map (with a RWMutex).
+type MemState struct {
```
timbray, line 27: RWMutex seems to have a bit of a bad rep. Maybe a few words of comment about why this, rather than for example figuring out a way to do an atomic update like smallTable does?

pruner/state.go
```diff
@@ -0,0 +1,81 @@
...
+		m: make(map[quamina.X]string),
+	}
+}
+
+var ErrExists = fmt.Errorf("pattern already exists for that X")
+
+func (s *MemState) Add(x quamina.X, pattern string) error {
+	s.lock.Lock()
+	var err error
+	if _, have := s.m[x]; have {
```
timbray, line 44: Hmm, interesting semantics here. Quamina Matcher won't stop you adding multiple different Pattern strings with the same X value, in fact that's how you can get an OR effect.  Which, to be fair, the README doesn't say, oops sorry.  I think pruner should be consistent with Quamina here?

timbray (changes requested):
Didn't really look at the unit tests much, assuming you have decent coverage.

Most of the comments are petty readability stuff.

One drop-dead blocker: As I didn't bother to tell anyone (sorry), you can call quamina.AddPattern(x, patternString) multiple times with the same x value and different patternStrings.  So you should allow multiple addPattern calls with the same x value.

pruner/README.md
jsmorph: The `MemState` is just a map of the live patterns, so there is no need to remember the identities of the ones that were deleted. But that is an implementation detail of that that one implementation of `State`, and those statements are confusing regardless. I'll reword. Maybe replace both statements with just

1. Remember the set of live patterns.

pruner/pruner.go
jsmorph: I have been accused of being overly fond of exceeding short, cryptic names (even symbols like δ).

In this case, the package itself sort of conveys (to me) what `Stats` and `Emitted` refer to.
`PatternStats` isn't quite right IMO since the data includes stuff not related to patterns (like `RebuildDuration`). The package is `pruner`, so `PrunerStats` seems redundant. 

`Live`, `Added`, `Emitted`, and `Filtered` all refer to patterns, which is a `Pruner`'s stock-in-trade, so adding the suffix `Pattern` to each also seems redundant.

All that said, I can suppress my character flaws for the greater good.

pruner/pruner.go
jsmorph: I'll try to add some gear to unit tests to verify that no count goes negative.

pruner/pruner.go
jsmorph: That's an interface method, which is documented [there](https://github.com/timbray/quamina/pull/4/files#diff-9dc1943e866b16c321a48f4c9314f8d547dc0807e08ec94b1ee6576291a20b17R149-R155). The type has a decent doc string. Does this method need one, too?

pruner/README.md
timbray: D'oh, you just filter anything that isn't in the patterns-I-know-about list. Don't mind me.

pruner/pruner.go
jsmorph: Hmm, I'm wondering if `Emitted` and `Filtered` should still be `int64` so overflow isn't feasible on any platform.

timbray: Oh, good catch.  Definitely.  Somebody is guaranteed to leave one of these things running for weeks.

timbray: where by weeks I mean years.

pruner/state.go
jsmorph: Even I'll admit that this name is a little vague. Hmm. `LiveSet`, `LivePatterns`, `LiveSetState`. `PrunerState` is both redundant (according to me) and wrong (since a Pruner has more state that that). Maybe `PatternState`. Yah, I know it seems like I'd argue against myself there. Or is it `PatternsState` anyway?

pruner/state.go
jsmorph: Yah, some quick commentary does seem helpful here. Since the application can provide its own `State` implementation, I was leaning in favor of simplicity here. [`sync.Map`](https://pkg.go.dev/sync#Map) seems more distasteful than `RWMutex`. I wanted a `map`-based implementation for simplicity -- or something similar that's built from the standard library. But I failed to come up with any better on balance. In particular, I don't see how to pull off some sort of atomic update of a field while keeping things simple.

FWIW, I recall some interesting discussion a few years back about what assignments were and were not considered atomic -- and on what platforms, etc. Official docs [here](https://go.dev/ref/mem), I guess. (Bringing out [`sync.atomic`](https://pkg.go.dev/sync/atomic) didn't obviously seem to help me anyway.)

Maybe some cleverness for this stock implementation can evolve.

pruner/state.go
jsmorph: Oh, wow. I should have checked that! Yeah, sure seems like we should be consistent here. Will fix.

pruner/pruner.go
jsmorph: After thinking about the type of the `Emitted` counter based on your comments, seems like we might in fact want `int64` here so that there is no cast in the arithmetic. Since an `Emitted int` could reasonably overflow on some platforms, seems like `Emitted` should be an `int64`. Keep these types consistent. (I'll resist arguing for an `uint64` I hope.)

pruner/pruner.go
jsmorph: Trying to stay with Go style that prefers very [short variable names](https://github.com/golang/go/wiki/CodeReviewComments#variable-names). But sure.

I missed this comment in my first pass.

pruner/pruner.go
jsmorph: Ah, I missed this comment, too. See above re [short variable name style](https://github.com/golang/go/wiki/CodeReviewComments#variable-names). But still yeah, I can do that.

pruner/pruner.go
jsmorph: Fixed I think when I changed `State.Get()` to `State.Contains()`.

pruner/state.go
jsmorph: Fixed.

pruner/state.go
```diff
@@ -0,0 +1,100 @@
...
+	_, have := s.m[x]
+	s.lock.RUnlock()
+	return have, nil
+}
+
+func (s *MemState) Delete(x quamina.X) (bool, error) {
+	s.lock.Lock()
+	_, had := s.m[x]
+	if had {
+		delete(s.m, x)
```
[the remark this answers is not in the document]
jsmorph: To keep an accurate count of the number of live patterns.

But I just realized that code is wrong because of the one-to-many ids-to-patterns capability!

Will fix.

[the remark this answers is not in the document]
jsmorph: PrunerMatcher.DeletePattern() doesn't need to return (bool,error) and could just return the error. Better to do it that way? Sometimes I like to know if my Delete found what I think it should find. Not a big deal.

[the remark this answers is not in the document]
jsmorph: See c4d3b4d for at least one fix. Happy to drop the bool from DeletePattern.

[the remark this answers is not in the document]
timbray: Yeah, if you don't mind I'd prefer to drop that boolean.  In fact, it's not obvious to me that DeletePattern has any useful return value at all.  How can an error occur?

But, you implicitly raise a good point.  Is there a case for an API to discover what patterns are current in a Matcher?  Probably not just yet, but let's keep it in mind.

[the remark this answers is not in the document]
timbray: Oh wait, DeletePattern(nil) is probably a bug, so I guess we should retain the error.

jsmorph:
On Wed, May 18, 2022 at 11:21 PM Tim Bray ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In pruner/state.go
> <https://github.com/timbray/quamina/pull/4#discussion_r876580499>:
>
> > +	s.lock.Unlock()
> +	return nil
> +}
> +
> +func (s *MemState) Contains(x quamina.X) (bool, error) {
> +	s.lock.RLock()
> +	_, have := s.m[x]
> +	s.lock.RUnlock()
> +	return have, nil
> +}
> +
> +func (s *MemState) Delete(x quamina.X) (bool, error) {
> +	s.lock.Lock()
> +	_, had := s.m[x]
> +	if had {
> +		delete(s.m, x)
>
> Yeah, if you don't mind I'd prefer to drop that boolean. In fact, it's not
> obvious to me that DeletePattern has any useful return value at all. How
> can an error occur?
>
Happy to drop the bool.

For the interface method State.Delete() in general, an error could occur
with an implementation that provides persistent state and therefore needs
to do I/O. Yes?

> But, you implicitly raise a good point. Is there a case for an API to
> discover what patterns are current in a Matcher? Probably not just yet, but
> let's keep it in mind.
>
Interesting. Yeah, I could imagine a use for such an API or APIs (like
Get(x) and Iterate(func(x,[]string)), which are close to what State can do.
State doesn't currently have Get(x). Only Contains(x))

--Jamie

pruner/pruner.go
```diff
@@ -0,0 +1,359 @@
...
+	rebuildTrigger rebuildTrigger
+
+	// lock protectes the pointer the underlying Matcher as well as stats.
+	//
+	// The Matcher pointer is updated after a successful Rebuild.
+	// Stats are updated by Add, Delete, and Rebuild.
+	lock sync.RWMutex
+}
+
+var defaultRebuildTrigger = newTooMuchFiltering(0.2, 1000)
```
jsmorph, line 85: That's the filtered/emitted min ratio and min number of total X's emitted.

What should these constants be?

pruner/pruner.go
timbray: Wait, won't this Matcher type conflict with the new Matcher interface?

pruner/pruner.go
timbray: *snicker* what makes you think I have a better idea. Might as well run with these.

pruner/pruner.go
timbray: sure

timbray (commented):
LGTM

timbray (approved):

Issue #6, opened 2022-05-24
Enable Github Actions


embano1:
As per a discussion with @timbray it would be nice to leverage Github Actions for same basic automation around unit tests, PR reviews, etc. Later this can be extended for release automation, package management (if needed), and so on.

embano1:
@timbray we can also do 1,2,3 and I rebase to keep the PR up2date. Either way works for me. After merging (3) we should enable branch protection rules to ensure PRs must pass certain status checks (actions) before being allowed to merge. (admins can always overrule though if needed)

embano1:
Btw: at some point (perhaps later) you might want nice CHANGELOGs or release notes. This typically requires some effort on commit naming (prefixing)/content. Just wanted to mention here as the more commits you create without a "standardized" (parsable) structure, the harder it gets later to create an automated CHANGELOG/release notes :)

There's multiple options here, let me know if this sounds useful and I can create a separate issue to discuss/raise a PR.

Example: https://github.com/vmware/govmomi/releases

timbray:
Oh yes please, definitely useful, a Issue/PR would be very welcome.

PR #7, opened 2022-05-24
Add Github Actions and Dependabot automation

embano1:
Changes:

- [x] Add [Dependabot](https://github.com/dependabot) configuration to automatically update any dependencies, incl. Github Actions
- [x] Enable Code Analysis (security) in PRs and push to `main`
- [x] Enable dependency review in PRs (for newly added deps, if any)
- [x] Add Go unit tests action in PRs and push to `main`
- [x] Add [golangci-lint](https://github.com/golangci/golangci-lint) action in PRs and push to `main`
- [x] Add close stale action to clean up stale issues/PRs

Notes:

- unit tests are failing (you're aware of this IMHO)
- golangci-lint tests are failing (see the errors in the code annotations)
- Dependency review needs to be enabled for the repo to work
- You might want to enable branch protection to require certain CI workflows to pass before allowing to merge a PR (see ["enable required status checks"](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule#creating-a-branch-protection-rule))

Closes: #6 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+      interval: "daily"
--- a/.github/workflows/dep-review.yaml
+++ b/.github/workflows/dep-review.yaml
@@ -1,0 +1,23 @@
+name: Dependency Review
+
+on:
+  pull_request:
+    branches: ["main"]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    strategy:
+      matrix:
+        go-version: ["1.18"]
+        platform: ["ubuntu-latest"]
+    runs-on: ${{ matrix.platform }}
+    timeout-minutes: 5
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        
+      - name: Dependency Review
+        uses: actions/dependency-review-action@a9c83d3af6b9031e20feba03b904645bb23d1dab
--- a/.github/workflows/go-lint.yaml
+++ b/.github/workflows/go-lint.yaml
@@ -1,0 +1,26 @@
+name: Code Linting
+
+on:
+  push:
+    branches: ["main"]
+
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  lint:
+    name: Code Linting
+    strategy:
+      matrix:
+        go-version: ["1.18"]
+        platform: ["ubuntu-latest"]
+    runs-on: ${{ matrix.platform }}
+    timeout-minutes: 5
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+      with:
+        fetch-depth: 1
+        
+    - name: Run golangci-lint
+      uses: golangci/golangci-lint-action@537aa1903e5d359d0b27dbc19ddd22c5087f3fbc
--- a/.github/workflows/go-unit-tests.yaml
+++ b/.github/workflows/go-unit-tests.yaml
@@ -1,0 +1,72 @@
+name: Unit Tests
+
+on:
+  push:
+    branches: ["main"]
+
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  test:
+    name: Unit Tests
+    strategy:
+      matrix:
+        go-version: ["1.18"]
+        platform: ["ubuntu-latest"]
+
+    runs-on: ${{ matrix.platform }}
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+
+      - name: Install tparse
+        run: go install github.com/mfridman/tparse@latest
+
+      - name: Set up Go ${{ matrix.go-version }}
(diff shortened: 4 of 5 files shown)


.github/workflows/stale.yaml
```diff
@@ -0,0 +1,31 @@
+name: Close Stale
```
timbray, line 1: I'm a little dubious about this one so early in the project's life.  It's not as though we are besieged with input, also I'm retired and Quamina is if not my primary hobby it's close to the top, and if we don't like a PR or an issue we should boot it or close it.

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
```
timbray, line 7: What's the effect of this if we get a flurry of little PRs?

.github/workflows/dep-review.yaml
```diff
@@ -0,0 +1,23 @@
+name: Dependency Review
```
timbray, line 1: As noted in conversation, I want to reduce Quamina's dependencies to as close to zero as possible.  For example, the event flattener has its own JSON parser which is insanely faster than Go's default one (which first of all isn't that hard and also Quamina's goals are very different).  Is there anything to do in this section to reflect that design goal?

.github/workflows/go-lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+    name: Code Linting
+    strategy:
+      matrix:
+        go-version: ["1.18"]
+        platform: ["ubuntu-latest"]
+    runs-on: ${{ matrix.platform }}
+    timeout-minutes: 5
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
```
timbray, line 21: What's this opaque string?  Just curious.

.github/workflows/go-unit-tests.yaml
```diff
@@ -0,0 +1,72 @@
...
+    branches: ["main"]
+
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  test:
+    name: Unit Tests
+    strategy:
+      matrix:
```
timbray, line 14: Sorry for requests from the uninitiated, but once again I'm curious: What is matrix?

embano1:
> Quick question: If a unit test is failing and it doesn't have "Benchmark" or "Concurrency" in the name, I *don't* know about that and would like to hear details.

Sorry, I am not understanding the question. Can you please elaborate further?

.github/workflows/stale.yaml
embano1: Sure, happy to remove this one for now!

.github/dependabot.yml
embano1: For now we should not hit this limit because you don't have any deps in Go and we just have a hand-ful of Github Actions which need maintenance. But yes, if there's too many deps which need to be updated at the same time, this limits the number of PRs the bot will create.

.github/workflows/dep-review.yaml
embano1: This action first helps to detect newly added dependencies and forces PR creators to keep them up2date.

So I think it's useful for your goal of (close to) 0 deps:

a) flagging/highlighting unwanted deps in PRs
b) as project grows, maintain healthy dependencies

.github/workflows/go-lint.yaml
embano1: best practice to use commit SHAs instead of branch names or tags to prevent abuse

.github/workflows/go-unit-tests.yaml
embano1: All good: matrix strategies are super powerful, even though here we don't use them (yet).

Matrix strategies help you define various combinations, e.g. different go versions, which you can use to templatize the action.

Reason why I go with matrix setup even though it just uses one target (setting), is that later it's super easy to add another setting. For example, imagine you want to unit test various Go runtimes, or different go test settings. This all can be input to the matrix test and Github Actions will spin up multiple (concurrent) test pipelines based on the matrix table.

.github/dependabot.yml
timbray: OK

.github/workflows/go-lint.yaml
timbray: Fair enough. [Googling](https://www.google.com/search?client=safari&rls=en&sxsrf=ALiCzsZCexC7UB6Y_ktzIAvWIIr4FO5YTg:1653493427591&q=actions/checkout@+%222541b1294d2704b0964813337f33b291d3f8596b%22&sa=X&ved=2ahUKEwixi4-D__r3AhVKiI4IHSgDABgQ5t4CegQIGxAB&biw=1725&bih=1015&dpr=2) reveals this is used a lot of places.  Pardon me but I'm curious, is there a way to use this as a link to see what is actually being referenced?

.github/workflows/go-unit-tests.yaml
timbray: OK, I get it, matrix is a keyword for this GitHub tool I haven't learned yet.  Another item for my to-do list.

embano1:
> Anyhow, so if there are unit-test failures outside those two cases I'd be surprised and concerned. Let's not worry now, I'll have a close look after everything's in place.

Gotcha. So we don't have to make unit tests required for PR merges. So you see the result and can still decide whether it's due to a known issue/flakes.

Once `tparse` gets an update, I'll file another minor PR so we get such nice output instead of raw terminal logs from the actions :)

image

PR #9, opened 2022-05-24
Package reorg: pkg and internal; closes #5; see #6

### LINKED ISSUE #5: go.mod should define the module name with complete github.com path
It seems that the way the module name is defined in go.mod, it is not possible to easily use this package from other golang projects.
Is there reason, that the module name is not defined as follows?

```
module github.com/timbray/quamina

go 1.18
```

Alternatively, if there is an easy way to use this package without this change, then could it be added to the README?

jsmorph:
1. `lib` &rarr; `pkg/core`
2. `pruner` &rarr; `pkg/pruner`
3. New `internal/tools` for some shared test tools (?)
4. `go.mod` also updated with full module name

Closes #5 and see #6.


jsmorph:
Re that vague unease: Yeah, I get it, but for me that unease might be offset by the potential benefits of having some simple tools that both `core` and `pruner` (and perhaps future things) can use. (I'm thinking in particular of future, shared performance test tools.) I'll admit that the use addressed here (getting test data) isn't compelling. I'm happy to drop it in favor of the simpler approach.

jsmorph:
Re big files: Maybe at some point the big files can be replaced with data that's carefully generated. Maybe not. The sketch [here](https://github.com/jsmorph/quamina/tree/gencorpus/internal/gen) aspires to generate good data in a flexible, controllable way.

timbray:
OK, I'm basically fine with this, but I'm out of touch with current Go best practices. @embano1 tells me there's a movement away from /pkg/ so I asked him to look at this PR and present alternatives.

embano1:
Couple questions:

Why don't we have a `go.sum` file? This should (must) be committed to the repo

`pkg/` prefix is considered an anti-pattern these days in Go (adds no value), i.e. just make it `core/`, `pruner/`, etc.

`internal/tools`: are these purely for testing? If so, no need to put them in `internal` and just make all tool files have `_test.go` so they cannot be imported (used) from outside the repo

embano1:
You might also want to rename `test_data` to `testdata` since `The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.` (https://pkg.go.dev/cmd/go/internal/test)

jsmorph:
> Why don't we have a `go.sum` file? This should (must) be committed to the repo

I'll add it here.

> `pkg/` prefix is considered an anti-pattern these days in Go (adds no value), i.e. just make it `core/`, `pruner/`, etc.

Good! I'll remove it here.

> Also, how do you envision the community using the package? Do we really need the `core/` folder? I.e. would it be ok to just do `quamina.NewFJ()`? Since this is a package-only repo, having the "core" logic just sitting in the root folder is totally acceptable.

I don't have a strong opinion. In my experience, I seem to end up wanting Go files in subdirectories, but I also agree that there might not be any good reason to do that now.


> `internal/tools`: are these purely for testing? If so, no need to put them in `internal` and just make all tool files have `_test.go` so they cannot be imported (used) from outside the repo

The problem is that two packages want to use these shared test tools.

However, @timbray expressed skepticism that having these particular shared tools is warranted, and I agree. For now at least I'll remove `internal` entirely. Simpler without sacrificing much in the way of reuse.

jsmorph:
> > The problem is that two packages want to use these shared test tools.
> 
> Is the use for unit testing (e.g. two packages calling this API from `_test.go` files)?

That was the intention, but I think it's not worth the trouble right now. So I'll removed `internal/tools` in the latest commits.

timbray:
@jsmorph this ready to go?  @embano1 looks OK to you now?

jsmorph:
> @jsmorph this ready to go?

Yeah, I think so. Only pending question I know is `core` directory versus those `*.go` in root of the repo. I left the `core` directory in place for now.

embano1:
LGTM, I'm ok with `core` for now since the API is considered unstable anyways.

Issue #34, opened 2022-05-30
api: Relationship between Flatteners and Matchers


timbray:
I'm worried about getting into one of those hideous Go circular-dependency traps.  Let me describe the situation in free text in hopes of clearing up my own thinking and provoking suggestions on the right way forward.

After a Matcher has been loaded with Patterns, every Event has to be flattened before it can be used with a Matcher call to find which patterns match it.

Both Matcher and Flattener are interfaces, both currently defined in core/.  We already have two implementations of Matcher (in core/ and pruner/) but only one implementation of Flattener (core/FJ).  I think it is very likely that we will have multiple flatteners in future - for Avro, Protobufs, CBOR, etc etc.  

In theory, Matchers and Flatteners don't need to know about each other. A Flattener gets a []byte message in some form and emits a []*Field. Hmm, just realized they both need to know about Field.  

A Flattener does need to know which Field names are used in patterns, so it can leave the rest out of the []Field output - this is a really important optimization.  In practice this has to be provided by a Matcher because it knows what fields are used as a result of all the AddPattern() calls.  There is an interface called NameTracker with one method, IsNameUsed(string name), which in theory should be used by FJ but at the moment it isn't, NewFJ() is given a CoreMatcher, which it uses both for the IsNameUsed call and for the FlattenAndMatch convenience function. So that interface needs to be either used or discarded.

I want to make it easy and straightforward to add a new Flattener for a new data type and now I'm thinking of how to organize things to achieve that. Conceivably the new Flattener shouldn't even have to be part of this repo.  Two approaches occur to me:

1. Leave everything in core/ - minimum hassle and risk of circular dependencies.  But, is it still straightforward to implement a new Flattener?
2. Make new subdirectories for at least Matcher, Flattener, Field.  You wouldn't get a convenience function like flattener.FlattenAndMatch() but everything would be maximally clean. I'd lose sleep about wandering into a circular dependency.

Anyhow, at least I've written the problem down now. Input at any level from comprehensive suggestions to "at least don't do X" would be welcome.

jsmorph:
Maybe:

1. package `matching`: interfaces like `Flattener` and `Matcher` (and maybe `NameTracker`) and some shared primitive types like `Field`.
2. package `core` for the core Matcher implementation.
3. `flatterners/json` for `FJ`. (Maybe not "json" for the package name.)
4. package `pruner` for the Pruner implementation -- pretty much as is.

[...]

PR #38, opened 2022-05-30
api: reorganize directory layout

timbray:
Closes: #34

This was created very hastily to drive discussion of good directory layouts.  The goal is to make it entirely easy to add new Flatteners to the system, and for callers to ignore all this stuff about flatteners and matchers, and to avoid circular dependencies.


timbray:
Upon further reflection, I think that as @embano1 said, we don't need a separate `pruner` directory, that stuff can go in with `core`.  I guess we also don’t need a separate `fields` and `constants`, can have a `shared`.  So that would leave us with

`control`
`core`
`flattener`
`shared`
`testdata`

PR #40, opened 2022-05-31
api: reorganize directory layout

timbray:
Closes: #34

Here is a reorganization of the repository with everything at the top level, as is said to be suitable for pure library projects. It's remarkably simple but wow, is that ever a lot of *.go files. I'm tempted to group them into directories (but still in the same 
package) just to make it a little more orderly to the eye.  I bet there will be loads of development on *_matcher.go but very little in future on FJ, so maybe put it in a rarely-opened subdirectory.  Anyhow, I'll sleep on this but for the moment I'm happy with it.

Hmmm, if we merge this, README.md would be totally out of date and wrong.


jsmorph:
@timbray, do you want to discuss `MatchesForFields` in the README?

timbray:
> @timbray, do you want to discuss `MatchesForFields` in the README?

Nope, it's no longer exported and external callers don't need it.  Obviously, it's accessible to code we write in the `quamina` package.

README.md
```diff
@@ -10,19 +10,21 @@
 
 ### Fast pattern-matching library
 
-Quamina provides APIs to create an interface called 
-a **Matcher**,
-add multiple **Patterns** to it, and then query JSON blobs
-called **Events** to discover which of the patterns match 
+**Quamina** implements a data type that has APIs to 
+create an instance and add multiple **Patterns** to it,
```
embano1, line 14: question: along the text, there's a mix of "Patterns" vs "patterns". Is this intentional?

README.md
```diff
@@ -10,19 +10,21 @@
...
 ### Fast pattern-matching library
 
-Quamina provides APIs to create an interface called 
-a **Matcher**,
-add multiple **Patterns** to it, and then query JSON blobs
-called **Events** to discover which of the patterns match 
+**Quamina** implements a data type that has APIs to 
+create an instance and add multiple **Patterns** to it, 
+and then query data objects called **Events** to
+discover  which of the patterns match
```
embano1, line 16: nit: whitespace

README.md
```diff
@@ -128,87 +130,79 @@ Number matching is weak - the number has to appear
...
 The first step in finding matches for an Event is 
 “flattening” it, which is to say turning it 
 into a list of pathname/value pairs called Fields. Quamina 
-defines a `Flattener` interface type and provides a 
-JSON-specific implementation in the `FJ` type.
+defines a `Flattener` interface type and has a built-in
+`Flattener` for JSON.
 
 `Flattener` implementations in general will have
 internal state and thus not be thread-safe.
```
embano1, line 146: nit: "thus not be thread-safe"? Since this is an interface, IMHO we cannot tell whether an implementation is/must be thread-safe.

README.md
```diff
@@ -128,87 +130,79 @@ Number matching is weak - the number has to appear
...
-	DeletePattern(x X) error
-}
-```
-
-Above are the operations provided by a Matcher. Quamina
-includes an implementation called `CoreMatcher` which
-implements `Matcher`.  In a forthcoming release it will
-provider alternate implementations that offer extra
-features.
+func New(...Option) (*Quamina, error)
```
embano1, line 160: nit: since you use argument names in the other functions, add `opts` here too.

README.md
```diff
@@ -128,87 +130,79 @@ Number matching is weak - the number has to appear
...
+func WithMediaType(mediaType string) Option
+func WithFlattener(f Flattener) Option
+func WithPatternDeletion(b bool) Option
+func WithPatternStorage(ps LivePatternsState) Option 
 ```
+For example:
+
 ```go
-func pruner.NewMatcher() *Matcher
+q, err := quamina.New(quamina.New(quamina.WithMediaType("application/json")))
```
embano1: why the double quamina.New(quamina.New())?

README.md
```diff
@@ -128,87 +130,79 @@ Number matching is weak - the number has to appear
...
-
-Create new Matchers, take no arguments. The difference
-is that the `pruner.NewMatcher` version supports the
-`DeletePattern()` API. Be careful: It occasionally
-rebuilds the Matcher in stop-the-world fashion, so if you
-delete lots of Patterns in a large Matcher you may 
-encounter occasional elevated latencies.
+The meanings of the `Option` functions are:
+
+`WithMediaType`: In the futue, Quamina will support
```
embano1, line 174: question: do we need that option/make it part of the API? I thought picking/implementing the corresponding `Flattener` would implicitly do that?

README.md
```diff
@@ -128,87 +130,79 @@ Number matching is weak - the number has to appear
...
+yet implemented.***
+
+### Data APIs
 
 ```go
-func (m *Matcher) AddPattern(x X, patternJSON string) error
+func (q *Quamina) AddPattern(x X, patternJSON string) error
 ```
-
 The first argument identifies the Pattern and will be
```
embano1, line 203: question: first or second argument? looks like it should be second, since `X` is the data?

README.md
```diff
@@ -128,87 +130,79 @@ Number matching is weak - the number has to appear
...
-returned by a Matcher when asked to match against Events.
-X is currently `interface{}`. Should it be a generic now
-that Go has them?
+returned by Quamina when asked to match against Events.
+X is defined as `any`.
 
 The Pattern must be provided as a string which is a
```
embano1, line 207: question: if I pick a different `Flattener`, why not call it `pattern`, i.e. make it not about JSON?

README.md
```diff
@@ -217,97 +211,99 @@ The `error` return is used to signal invalid Pattern
...
-`FJ.Flatten` and `Matcher.MatchesForFields` APIs.
+
+This generates a copy of the target instance. Such
+copies may safely run in parallel in different
+goroutines executing any combination of
+`MatchesForEvent()`, `AddPattern()`, and
+`DeletePattern()` calls.  There is a significant
+performance penalty if there is a high rate of
+`AddPattern()` invocations in parallel with
+`MatchesForEvent()`.
```
embano1, line 256: question: above you say a single instance is not thread-safe. Here you write about parallel operations (I read that as concurrent operations). At least, this is confusing as the reader might not be clear about which operations/objects are thread-safe or not :)

README.md
```diff
@@ -217,97 +211,99 @@ The `error` return is used to signal invalid Pattern
...
+unique fields that appear in all the patterns that have
+been added to Quamina, but still remains sublinear
+in that number.
 
 A word of explanation: Quamina compiles the
-patterns into a somewhat-decorated automaton and uses 
-that to find matches in events; the matching process is 
+patterns into a somewhat-decorated automaton and uses
+that to find matches in events; the matching process is
 O(1) in the number of patterns.
```
embano1, line 282: nit: use syntax highlighting for O(1) as above

README.md
```diff
@@ -217,97 +211,99 @@ The `error` return is used to signal invalid Pattern
...
 However, for this to work, the incoming event must be
-flattened into a list of pathname/value pairs and 
-sorted.  This process exceeds 50% of execution time, 
+flattened into a list of pathname/value pairs and
+sorted.  This process exceeds 50% of execution time,
 and is optimized by discarding any fields that
 do not appear in one or more of the patterns added
-to the matcher. Thus, adding a new pattern that only
+to Quamina. Thus, adding a new pattern that only
 mentions fields mentioned in previous patterns is
```
embano1, line 290: question: "mentions fields mentioned"?

README.md
timbray: I was trying to be principled and capitalize when I was using it to specifically mean the Quamina type.  But there's been a lot of editing and it's possible there are breakages.

README.md
timbray: I said "in general will have internal state" - it's possible you could have a thread-safe flattener but I don't think the language rules it out.

README.md
timbray: Patched since you read this

README.md
timbray: No, the X identifies the pattern, the pattern string is the pattern-language specification of it

README.md
timbray: Haha, you're talking to an IETF guy.  The correct way to identify byte blobs crossing the Internet is with media types, so Flatteners should be keyed by those.  If you want to bring your own Flattener to support your private secret format, we support that.  But for something you got from an HTTP server, this is a good approach.

README.md
timbray: well, I mean if you have a bunch of goroutines making Quamina API calls, each with a different Quamina instance, if a high proportion of those calls are AddPattern, performance will suffer.   OK, will try to clarify.

README.md
embano1: Got it. But in "thus not be thread-safe" there's still something missing, e.g. "thus are not thread-safe"?

README.md
embano1: I knew you came from that direction :)

There's just a problem here with the options then. Will explain in the corresponding code section. IMHO `Flatteners` should be keyed by mediatype, e.g. we own the defaultMatcher which is keyed by JSON. If someone wants to overwrite, they likely know the mediatype - hence my question why have two APIs basically doing the same thing?

README.md
timbray: Oops, need to be clearer. Right now we should be able to support Events in lots of different formats, once there are Flatteners.  But Patterns are just in JSON. I suppose in principle you could also support Patterns in protobufs etc but that wouldn't work well because Patterns are designed to be human-authored and humans can't read Avro. So for the moment JSON will do.

quamina.go
```diff
@@ -0,0 +1,94 @@
...
+}
+func WithFlattener(f Flattener) Option {
+	return func(q *Quamina) error {
+		if f == nil {
+			return errors.New("nil Flattener")
+		}
+		q.flattener = f
+		return nil
+	}
+}
```
embano1, line 39: There's an apply issue with these two options competing for the final outcome and I think those options are either redundant or mutual exclusive - the latter makes it harder with the function options pattern. Let me explain.

Since those are options, we have no control over the order of how a user passes these options. Neither should a user care. However, the following can happen:

`opts := []quamina.Option{WithFlattener(myCustomJSONOne), WithMediaType("application/JSON")}`

In this case, the user would not get the desired `Flattener` but our default `newJSONFlattener`.

My proposal is to drop one of these options, preferably the `MediaType`. Reason:

- The `Flattener` implicitly deals with a specific media type - why do I have to specify one?
- If we drop `WithFlattener`, users lose control over the applied `Flattener` and we'd have to constantly update `WithMediaType` `switch` statement
- If we drop `WithMediaType`, the user gets our default (JSON) or can overwrite the `Flattener` as needed and use a custom (or also JSON) media type based on the implemented `Flattener`

Since `Flattener` is an interface, we don't even have to own these in our repo. A user can "bring" it and does not have to care about the semantics of also calling `WithMediaType`.

Not sure if I'm understanding the rational behind `WithMediaType` well. Looks like it's purely for `Flattener` selection which complicates the functional option pattern a lot. For me the media type behavior is implicit and the `Flattener` would throw an error if it gets a mismatched event anyways.

README.md
```diff
@@ -217,97 +211,99 @@ The `error` return is used to signal invalid Pattern
...
-Transforms an event, which must be JSON object
-encoded in UTF-8, into a list of `Field` instances.
+A single Quamina instance can not safely be used by
+multiple goroutines at the same time.  However, the
+underlying data structure is designed for concurrent
+access and the `Copy` API is provided to support this.
 
 ```go
-func (fj *FJ) FlattenAndMatch([]byte event) ([]X, error)
+func (q *Quamina) Copy() *Quamina
```
jsmorph, line 246: Is "Copy" the right name here? The code doesn't actually copy anything. (Same goes for `Flattener.Copy()`, I guess.) The word "copy" (along with not having read the actual implementation!) threw me off initially. Makes me think of a deep copy.

If I complain about "Copy", I should propose an alternative. Sadly I don't yet have a good one. In my mind it's something like `MatchingSession`, but that sounds terrible.

quamina.go
timbray: I think on this one we'll have to disagree.  The Internet has a well-established, well-debugged, broadly implemented standard for identifying data formats, so if we're going to identify data formats, we should use it. I think we still need WithFlattener for custom Flatteners that aren't associated with a well-known media type or offer some nonstandard custom feature. I agree that specifying both WithFlattener() and WithMediaType() doesn't make sense, so I'll put a check in the code to throw an error. Also doesn't make sense to specify both WithPatternDeletion(true) and WithPatternDeletion(false).

PR #45, opened 2022-06-02
fix: Minor fixes in docs and style

embano1:
Ran `gofumpt -w .` to force style.

The comments were off (your editor seems to add two spaces in subsequent comments - this throws off the godoc interpreter. Examples/explanation:

```go
// this is a broken
//  comment because only one space allowed after //

// this is a good comment because two spaces will render syntax highlighting
//  {"rendered":"withhighlighting"}

// this is a normal and 
// good comment
```

You can check how the stuff gets rendered with:

```shell
go install golang.org/x/tools/cmd/godoc@latest

# from within your project where go.mod lives
godoc -http :6060 # open page and click on your project
```

I left some `TODO` in certain doc strings for exposed types/functions which need to be added/clarified.

### DIFF
--- a/arrays_test.go
+++ b/arrays_test.go
@@ -70,7 +70,6 @@ const bands = `{
 }`
 
 func TestArrayCorrectness(t *testing.T) {
-
 	// only pattern3 should match
 	pattern1 := `{"bands": { "members": { "given": [ "Mick" ], "surname": [ "Strummer" ] } } }`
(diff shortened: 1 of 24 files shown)


timbray:
It's not my editor, I like to extent lines in long descriptions to achieve a paragraph-like effect and improve readability for humans.  But if it's nonstandard, fine.  Nice tool; I'm surprised that a few of those things weren't caught by the linter.

Issue #61, opened 2022-06-07
api: Implement anything-but pattern


timbray:
(compatible with EventBridge)

PR #64, opened 2022-06-10
api: implement anything-but

timbray:
Closes: #61

(also includes some misc cleanup that obscures the diffs a bit, sorry)


timbray:
It'd be nice to get a review on this but it seems that nobody else has become sufficiently deranged enough to dive into the automaton-wrangling code.

Issue #70, opened 2022-06-16
Add example for godoc


embano1:
Create testable example which also shows up in godoc/https://pkg.go.dev/

PR #71, opened 2022-06-16
Add godoc example

embano1:
Add a `go test`-able example which also shows up in godoc and https://pkg.go.dev/ and also allows user to run ("interact") with the code.

Closes: #70

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,0 +1,54 @@
+package quamina_test
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/timbray/quamina"
+)
+
+const userRegisteredEvent = `{
+  "id": "1c0e1ce4-3d88-4786-a09d-7133c170d02a",
+  "type": "UserRegistered",
+  "user": {
+    "name": "Doe, John",
+    "premiumAccount": true
+  }
+}
+`
+
+const premiumUserPattern = `{
+  "type":["UserRegistered"],
+  "user": {"premiumAccount": [true]}
+}`
+
+func ExampleNew() {
+	q, err := quamina.New()
+	if err != nil {
+		log.Fatalf("could not create quamina instance: %v", err)
+	}
+
+	const patternName = "premium user"
+	err = q.AddPattern(patternName, premiumUserPattern)
+	if err != nil {
+		log.Fatalf("could not add pattern: %v", err)
+	}
+
+	matches, err := q.MatchesForEvent([]byte(userRegisteredEvent))
+	if err != nil {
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		log.Fatalf("could not create quamina instance: %v", err)
+	}
+
+	err = q.AddPattern("premium user", premiumUserPattern)
+	if err != nil {
+		log.Fatalf("could not add pattern: %v", err)
+	}
+
+	event, err := q.MatchesForEvent([]byte(userRegisteredEvent))
```
timbray, line 36: not crazy about the variable name "event" - what you get back is []X, so a better name would be "matches" or something like that.

example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		log.Fatalf("could not match for event: %v", err)
+	}
+
+	if event != nil {
```
timbray, line 41: I'm pretty sure that Quamina never returns nil. It always returns []X which of course might be empty.  The kind of check you want to do is something like 
```go
if len(matches) == 0 || matches[0] != "premium user" {
  t.Error(whatever)
}
```

timbray (changes requested):
This is useful because it reveals that that the existing documentation did not properly explain the return value from `MatchesForEvent`.

example_test.go
embano1: Ah, of course! Will do.

example_test.go
embano1: Ah! It's even in the docs. I just followed go "idioms" here, but you're right: it preallocates a slice!

timbray: Yes, that makes the calling sequence a little simpler for the caller, no need to check for nil.

example_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	for _, m := range matches {
+		if m == patternName {
+			fmt.Printf("pattern matched for event: %q", patternName)
+			return
+		}
+	}
+
+	fmt.Println("no pattern match")
```
embano1, line 49: @timbray I think this is even better/more clear now

example_test.go
embano1: I slightly improved it for better readability/to follow along the example.

timbray:
That's weird, slight decrease in coverage?  That doesn't make any sense to me, does it to you?

embano1:
> That's weird, slight decrease in coverage?  That doesn't make any sense to me, does it to you?

No, no idea...

PR #104, opened 2022-09-22
Bug Fix: exists:false failures are not pruned from match set depending on order

kylemcc:
First, thank you for the excellent library! I believe I've encountered a somewhat sneaky bug.

In short, presumed `exists:false` matches that have been invalidated are not pruned from the match set unless the path of the `exists:false` field comes last after sorting. The proposals are processed in reverse order, and unless the exists:false field is processed first, it is added to the match set [here](https://github.com/timbray/quamina/blob/cbaf72026561bcb8fd13f783a2d22ca8b7e5b547/core_matcher.go#L181), but is never removed.

I've implemented a simple test case [here](https://github.com/timbray/quamina/commit/6474b35e795ad4f4bfa9cde326a64f9771e29341) that fails on `main`.

The fix I've implemented is to remove the items in `failedExistsFalseMatches` from the `matches` set before returning. I'm not sure if this is the optimal fix, so let me know if there's a better way to achieve this.

### DIFF
--- a/core_matcher.go
+++ b/core_matcher.go
@@ -200,6 +200,9 @@ func (m *coreMatcher) matchesForSortedFields(fields []Field) *matchSet {
 			matches = matches.addX(presumedExistsFalseMatch)
 		}
 	}
+
+	// prune any invalidated exists:false matches from the set
+	matches = matches.removeX(failedExistsFalseMatches.matches()...)
 	return matches
 }
 
--- a/core_matcher_test.go
+++ b/core_matcher_test.go
@@ -122,6 +122,37 @@ func TestExerciseMatching(t *testing.T) {
 	fmt.Println(matcherStats(m))
 }
 
+func TestExistsFalseOrder(t *testing.T) {
+	j := `{
+		"aField": "a",
+		"bField": "b",
+		"cField": "c"
+	}`
+
+	// make sure exists:false properly disqualifies a match regardless of where
+	// it occurs (lexicographically) in the pattern
+	shouldNotMatches := []string{
+		`{ "aField": [ { "exists": false } ], "bField": [ "b" ], "cField": [ "c" ] }`,
+		`{ "aField": [ "a" ], "bField": [ { "exists": false } ], "cField": [ "c" ] }`,
+		`{ "aField": [ "a" ], "bField": [ "b" ], "cField": [ { "exists": false } ] }`,
+	}
+
+	for i, shouldNot := range shouldNotMatches {
+		m := newCoreMatcher()
+		err := m.addPattern(fmt.Sprintf("should NOT %d", i), shouldNot)
+		if err != nil {
+			t.Error("addPattern: " + shouldNot + ": " + err.Error())
+		}
+		matches, err := m.matchesForJSONEvent([]byte(j))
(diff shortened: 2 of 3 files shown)


timbray:
Love to see a PR with attention to testing.  Yep, I agree there's a bug, looks like your fix works. Agree, not obvious if there's a more optimal approach, gimme a day to have a closer look.

timbray:
I've dug into this and unfortunately it's way more complicated than I thought. I apologize in advance for this wall of text, it's me thinking out loud. 

I've pretty well convinced myself that the basic design of `{"exists": false}` is broken and needs reworking.

## How it works now

Suppose you add a pattern: 
```go
pattern := `{ "foo": [ { "exists": false } ] }`
q.AddPattern("p", pattern)
```

Quamina builds a machine that looks like this.

```go
coreMatcher {
    presumedExistFalseMatches: &matchSet { "P" },
    state: &fieldMatcher { 
      transitions: map[string]*valueMatcher { 
        "foo": &valueMatcher { 
          existsTransitions: []*fieldMatcher {
            &fieldMatcher { 
              existsFalseFailures: &matchSet { "P" },
            }
          }
        }
      }
    }
  }
}
```

In English: The matcher has a `presumedExistFalseMatches` containing `"P"`. Its start state transitions on `"foo"` to a valueMatcher that doesn't have a byte-level matcher, just `existTransitions` transitioning to another state that has an `existFalseFailures` for `"P"`.

So when we're trying to match `{"foo": 2}`, the matcher transitions on `"foo"` to the valueMatcher, then on the `existsTransition` to the second fieldMatcher, and at that point notices the `existFalseFailures` value.

The problem is that the `addPattern` code, when it's dealing with both `exists:true` and `exists:false` transitions, adds an entry to the valueMatcher's `existsTransitions` transitioning to a fieldMatcher. So you have to make the transition in order that you have invalidated an `exists:false`. This is why the pass/fail on `exists:false` is different depending on whether the field is lexicographically the last in the pattern; will skip the explanation on that for now.

What I provisionally think needs to be done is have the `existsFalseFailures` match values in the **source** state, and when inserting an `exists:false` pattern, don't create a new fieldMatcher state to transition to. This is going to be harder than it looks because currently the `addPattern()` logic assumes that for each field in the pattern you’re going to have a new state to branch to. 

And I think it's harder than that. I'm pretty sure that both the `presumedExistFalseMatches` and `existFalseFailures` need to track, not just the `X` match-labeling value, but the combination of field name and value. Currently just an intuition without an explanation. 

Also: The unit testing around `exists:false` in Quamina is obviously totally inadequate or this bug wouldn't have crept in.

kylemcc:
Thanks for the detailed response. This may be a bit beyond my capacity to properly fix. But, this library is useful to me, so happy to make an effort. Maybe expanding the test suite would be a good initial step.

timbray:
> Thanks for the detailed response. This may be a bit beyond my capacity to properly fix. But, this library is useful to me, so happy to make an effort. Maybe expanding the test suite would be a good initial step.

Yeah, I think I have to tackle this one.  If you wanted to make a PR with some unit tests that highlight the bug, that would be useful.

PR #109, opened 2022-09-26
matchSet.addX / matchSet.addXSingleThreaded performance improvements

yosiat:
- addX: adding length check to avoid copying the map if nothing was changed.
- Introducing addXSingleThreaded which allows update-in-place in single-threaded usages.

```
name \ time/op                          benchs/1_baseline  benchs/2_len_check  benchs/3_thread_safe_add  benchs/4_thread_safe_len_check
_Quamina-10                                   48.0µs ± 0%         33.9µs ± 0%                6.6µs ± 0%                      6.8µs ± 0%

name \ alloc/op                         benchs/1_baseline  benchs/2_len_check  benchs/3_thread_safe_add  benchs/4_thread_safe_len_check
_Quamina-10                                   41.0kB ± 0%         30.4kB ± 0%                8.9kB ± 0%                      8.9kB ± 0%

name \ allocs/op                        benchs/1_baseline  benchs/2_len_check  benchs/3_thread_safe_add  benchs/4_thread_safe_len_check
_Quamina-10                                      161 ± 0%            129 ± 0%                   28 ± 0%                         28 ± 0%
```

* `benchs/1_baseline` - baseline
* `benchs/2_len_check` - adding early returrn with length check to `addX` (still being used in matching)
* `benchs/3_thread_safe_add` - using `addXSingleThreaded` in matching
* `benchs/4_thread_safe_len_check` - adding early return with length check to `addXSingleThreaded` .. didn't improved the situation, which makes sense in this case.

### DIFF
--- a/flatten_json_test.go
+++ b/flatten_json_test.go
@@ -1,7 +1,7 @@
 package quamina
 
 import (
-	"io/ioutil"
+	"os"
 	"testing"
 )
 
@@ -123,7 +123,7 @@ func TestMinimal(t *testing.T) {
 }
 
 func testTrackerSelection(fj Flattener, tracker NameTracker, label string, filename string, wantedPaths []string, wantedVals []string, t *testing.T) {
-	event, err := ioutil.ReadFile(filename)
+	event, err := os.ReadFile(filename)
 	if err != nil {
 		t.Error(filename + ": " + err.Error())
 	}
--- a/core_matcher.go
+++ b/core_matcher.go
@@ -178,11 +178,11 @@ func (m *coreMatcher) matchesForSortedFields(fields []Field) *matchSet {
 		for _, nextState := range nextStates {
 
 			// if arriving at this state means we've matched one or more patterns, record that fact
-			matches = matches.addX(nextState.fields().matches...)
+			matches = matches.addXSingleThreaded(nextState.fields().matches...)
 
 			// have we invalidated a presumed exists:false pattern?
 			for existsMatch := range nextState.fields().existsFalseFailures.set {
-				failedExistsFalseMatches = failedExistsFalseMatches.addX(existsMatch)
+				failedExistsFalseMatches = failedExistsFalseMatches.addXSingleThreaded(existsMatch)
 			}
 
 			// for each state we've transitioned to, give each subsequent field a chance to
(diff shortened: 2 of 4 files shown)


flatten_json_test.go
```diff
@@ -123,7 +123,7 @@ func TestMinimal(t *testing.T) {
 }
 
 func testTrackerSelection(fj Flattener, tracker NameTracker, label string, filename string, wantedPaths []string, wantedVals []string, t *testing.T) {
-	event, err := ioutil.ReadFile(filename)
+	event, err := os.ReadFile(filename)
```
yosiat, line 126: Fixing deprecation warning - 
```
flatten_json_test.go:4:2: SA1019: "io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
```

timbray (changes requested):
if I look in the CI code-coverage report at https://app.codecov.io/github/timbray/quamina/commit/975b876cc65b5ad2e739392e22638c6f63e5f4ad I see that `match_set.go` coverage decreased noticeably, which I guess means the unit tests somehow now don't visit a lot of the lines here. That's weird and surprising, could you have a look and see what's going on?

yosiat:
There are two flows which aren't covered:
1. `addX` called with empty exes
2. `addX` called with existing set and we are adding exes

This can come up from the flow of exists false case, If I would add another "exists false" example to this test - `TestExerciseMatching` it will take care of number 2.

I see there are no direct unit tests for `matchSet`, I can add direct tests and this will fix up the coverage issue.
Is that desired? or we want to come-up with new tests for "exists false" that will cover this case?

timbray:
> I see there are no direct unit tests for `matchSet`, I can add direct tests and this will fix up the coverage issue. Is that desired? or we want to come-up with new tests for "exists false" that will cover this case?

Up to you, I'm sure you'll do the right thing.  I just hate to see any check-in which reduces coverage.

yosiat:
Added tests for match sets, I can't see the CI results because they are "awaiting for approval", but locally match set is 100% coverage with those specific tests.

PR #110, opened 2022-09-26
chore: enable "gofmt" in golangci

yosiat:

PR #112, opened 2022-09-27
chore: Update golangci-lint config

embano1:
Adds more linters to the `golangci-lint` action:

```yaml
linters:
  disable-all: false
  # https://golangci-lint.run/usage/linters/#enabled-by-default
  enable:
    - gofmt
    - bodyclose
    - errname
    - errorlint
    - exportloopref
    - gochecknoinits
    - goimports
    - goprintffuncname
    - gosec
    # - ireturn
    - misspell
    - nilerr
    - nilnil
    - predeclared
    - stylecheck
    - thelper
    - tparallel
    - unparam
    # - wrapcheck
    - whitespace
    # wastedassign is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649
    # - wastedassign
```

Closes: #111
Supersedes: #110 
Signed-off-by: Michael Gasch <15986659+embano1@users.noreply.github.com>


timbray (commented):
Hey Micha, his all looks good to me, but one question: Is there any downside?

embano1:
> Hey Micha, his all looks good to me, but one question: Is there any downside?

Don't think so, in fact the changes I had to make to the existing code base shows that we might have some subtle bugs in error handling and also inconsistencies in certain LoC, e.g. comments and `if 0 == len(something)`.

The action is set up to not check on existing files though, i.e. only new contributions will be vetted against it.

timbray:
> The action is set up to not check on existing files though, i.e. only new contributions will be vetted against it.

Wait, so the first time I check in foo.go it would get gofmt and other listing, but if I commit an update to it that doesn't happen?  Why wouldn't we unconditionally gofmt everything?

timbray:
Hey Micha, I just merged Yosi's PR, you're probably going to have to rebase, sorry.

flatten_json_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 )
 
-func bequal(a []byte, b []byte) bool {
+func equal(a []byte, b []byte) bool {
```
yosiat, line 8: Instead of using this method, we can use golang `bytes.Equal` (https://pkg.go.dev/bytes#Equal)

If you don't want to change in this PR, I can open another one.

timbray: Yes please, let's not mix up CI PRs with code PRs.  Thanks in advance.

Issue #113, opened 2022-09-28
Make the flattener smarter about skipping fields


yosiat:
Continuing the discussion from this issue: https://github.com/timbray/quamina/issues/108

Currently the flattener is fast and performs well on cases where the searched fields (from the patterns) are covering most of the event payload, in cases of large payloads and the searched fields are small subset of it, the flattener can be improved.

## Proposed Solution and POC results
My proposed solution consists of maintaining paths index being in use (as a tree data structure) and POC flattener written using Jx.

The source resides here - https://github.com/yosiat/quamina/tree/flat-hack inside "paths.go" and "jx.go", it passes most of the existing flattener tests except the error handling.

Benchmark results:
```
goos: darwin
goarch: arm64
pkg: github.com/yosiat/quamina-flatenner
Benchmark_Quamina_JxFlattener-10                 2612365               437.9 ns/op            24 B/op          3 allocs/op
Benchmark_Quamina_Flattener-10                     83246             14347 ns/op             360 B/op         23 allocs/op
Benchmark_LargePayload_QuaminaFlattner-10           78272             15307 ns/op            1240 B/op         44 allocs/op
Benchmark_LargePayload_JxFlattner-10               876013              1365 ns/op             904 B/op         24 allocs/op
PASS
ok      github.com/yosiat/quamina-flatenner     5.727s
```
source: https://github.com/timbray/quamina/issues/108#issuecomment-1258625539
Those benchmarks can't be shared currently since they are using an internal data, but can be adjusted and shared if needed.

BigShellStyle benchmark (in `benchmarks_test.go`) -
```
# Quamina
Field matchers: 27 (avg size 1.000, max 1), Value matchers: 1, SmallTables 54 (avg size 15.500, max 28), singletons 0
428,547.72 matches/second with letter patterns

# Jx
Field matchers: 27 (avg size 1.000, max 1), Value matchers: 1, SmallTables 54 (avg size 15.500, max 28), singletons 0
1,367,947.02 matches/second with letter patterns
```

And in CityLots benchmarks where we are pattern matching against the whole event data, we see there is no difference -
```
=== RUN   TestCityLots
Field matchers: 7 (avg size 1.500, max 3), Value matchers: 6, SmallTables 0 (avg size n/a, max 0), singletons 6

173,434.09 matches/second

--- PASS: TestCityLots (1.64s)
=== RUN   TestCityLots_JX
[...]

Issue #30, opened 2022-05-28
Organization of tests


timbray:
We have a lot of tests and some of them really aren’t “unit tests” but focused on coverage and concurrency, etc. Speaking as a developer, here are the things I want to do.

1. Run unit tests in the directory I'm working in, the equivalent of `go test`.
2. Run concurrency tests across all the directories. This is important since thread-safety is an important and advertised feature.
3. Coverage tests, to support our coverage badge
4. Linting tests
5. Acceptance tests, including all of the above, when you think you’re ready to do a PR

I suppose the main activities would be (1), (2), and (3,4,5)

Notes.
* For (1) we absolutely need to run in <10sec, so that a developer can run them every time they save a tiny change
* (1) should include one benchmark test to catch performance regressions, which can easily creep in even on a minor change. I think `TestCityLots` is a good choice since it explores worst-case performance of the Flattener and Matcher.
* The `-race` flag really slows things down, in particular the concurrency tests.  I don't think it's really needed in case (1)?

If it were 20 years ago, I think I’d say “we need a Makefile”.  Um… I think we need a Makefile?

embano1:
I haven‘t done any profiling why `-race` is slowing the tests down so massively. But also without `-race` unit tests should finish within <10s, agreed.

My pov: there should be no different tests for unit/concurrency/coverage. They‘re just flags to the go tool and should finish within a short time with a simple command, e.g. `make unit-test`.

The `-race` detector in Go is useful and does a good job of spotting general concurrency issues, e.g. from contributors new to the code base/new to Go.

(4) would be `make lint`. (5) would be `make tests` (combined target).

Example: this `Makefile` is simple and has a `make help` target for pretty-printing: https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/development/vmware-event-router/Makefile

timbray:
Haha, you're OK with the idea of a Makefile? I assumed that was too old-fashioned for a modern project.  I’ll do a Makefile PR.

On the subject of -race, it's 100% toxic in combination with the coverage flag, but probably isn't needed with that tool. It seriously slows down a couple of the Concurrency-specific tests, but also it's probably really adding value there.

Actually, I'm optimistic that we won't have concurrency problems going forward because the places to update the data structure have been pushed down into a very small number of atomic.Load/Store operations hidden behind update() functions. But I've been wrong before on concurrency stuff.

embano1:
[...]

PR #119, opened 2022-10-04
chore: flattenJson: efficient strings, arrays and objects skip. 

yosiat:
Implement a way to efficiently skip values, in this way we are just scanning the buffer until we find the end of string/array/object, and don't check inner values.

```
benchmark                                                old ns/op     new ns/op     delta
Benchmark_JsonFlattener_ContextFields-10                 17312         9053          -47.71%
Benchmark_JsonFlattener_MiddleNestedField-10             17426         10713         -38.52%
Benchmark_JsonFlattener_LastField-10                     17060         9553          -44.00%
Benchmark_JsonFlattner_Evaluate_ContextFields-10         17645         9495          -46.19%
Benchmark_JsonFlattner_Evaluate_MiddleNestedField-10     17935         11152         -37.82%
Benchmark_JsonFlattner_Evaluate_LastField-10             17475         9882          -43.45%

benchmark                                                old allocs     new allocs     delta
Benchmark_JsonFlattener_ContextFields-10                 23             3              -86.96%
Benchmark_JsonFlattener_MiddleNestedField-10             24             4              -83.33%
Benchmark_JsonFlattener_LastField-10                     22             2              -90.91%
Benchmark_JsonFlattner_Evaluate_ContextFields-10         33             13             -60.61%
Benchmark_JsonFlattner_Evaluate_MiddleNestedField-10     33             13             -60.61%
Benchmark_JsonFlattner_Evaluate_LastField-10             29             9              -68.97%

benchmark                                                old bytes     new bytes     delta
Benchmark_JsonFlattener_ContextFields-10                 560           48            -91.43%
Benchmark_JsonFlattener_MiddleNestedField-10             576           64            -88.89%
Benchmark_JsonFlattener_LastField-10                     544           32            -94.12%
Benchmark_JsonFlattner_Evaluate_ContextFields-10         992           480           -51.61%
Benchmark_JsonFlattner_Evaluate_MiddleNestedField-10     1000          488           -51.20%
Benchmark_JsonFlattner_Evaluate_LastField-10             824           312           -62.14%
```

See discussion here: https://github.com/timbray/quamina/issues/113#issuecomment-1266681752


flattener_bench_test.go
```diff
@@ -0,0 +1,156 @@
+package quamina
```
yosiat, line 1: Add the benchmarks I am presenting, I think it will be good to commit those (if you agree I'll move `status.json` to `testdata` directory)

What do you think?

timbray: +1 yes please.

flatten_json.go
```diff
@@ -475,6 +496,60 @@ func (fj *flattenJSON) readLiteral(literal []byte) ([]byte, error) {
 	return literal, nil
 }
 
+func (fj *flattenJSON) skipBlock(openSymbol byte, closeSymbol byte) error {
```
timbray, line 499: This func makes me nervous.  Normally when we're going through the buffer we use `fj.step()` and `fj.ch()`, is there a reason why this is different?

Also, nesting;   Suppose you have an array (arguments `[`, `]`) and one of the elements is an object, possibly deeply nested? Also, how about an object (`{`, `}`) and a member value is an array.

flatten_json.go
```diff
@@ -475,6 +496,60 @@ func (fj *flattenJSON) readLiteral(literal []byte) ([]byte, error) {
...
+		return fj.error("event truncated in mid-string")
+	}
+
+	i := 0
+	data := fj.event[fj.eventIndex:]
+
+	for i < len(data) {
+		c := data[i]
+
+		if c == '\\' && i+1 < len(data) && data[i+1] == '"' {
```
timbray, line 538: I'd put a short comment here explaining why you can skip all the other `\\` escapes, because the only thing you care about is an unescaped `"`

flattener_bench_test.go
yosiat: do we want to run the benchmarks in CI? I am not sure about it, since I don't github actions will give us good results and will take some time to run the benchmarks.

timbray (changes requested):
Nice! My only worry is about `skipBlock` but maybe I'm missing something obvious?

I notice in the coverage report that the coverage went down a bit: https://app.codecov.io/github/timbray/quamina/commit/9b8031b158a40424d90574f38d4993e140c67b48

I'm not a fanatic about coverage percent but when it goes down that always makes me nervous.  I looked at the actual coverage red/green and didn't see anything uncovered that looked very worrying, but please have a glance at the coverage and see if there's anything that surprises you.

flattener_bench_test.go
timbray: See issue #30 - one problem is that GitHub runs actions single-threaded on a pathetic tiny instance. So we probably can't routinely run them all.  This bothers me a bit because I want to make sure that nothing gets checked in that accidentally causes a noticeable slowdown.

flatten_json.go
yosiat: > Also, nesting; Suppose you have an array (arguments [, ]) and one of the elements is an object, possibly deeply nested? Also, how about an object ({, }) and a member value is an array.

Nesting is covered here with the levels, you can see that we are skeeping most of `status.json` which contains nested objects with arrays.

I'll add specific tests for that, to validate skipping of objects & arrays.

> Normally when we're going through the buffer we use fj.step() and fj.ch(), is there a reason why this is different?

This is actually an interesting case! I wrote the code initially without the `step` and `ch` and when I used those methods the performance (time taken, not allocations) increased, talking about something like:
```
for {
  ch := fj.ch()

  // process

  if fj.step() != nil { return err }
}
```

So I tried to investigate this, it didn't make any sense, since this methods are inlined and cheap, so I did manual inlining and got no improvements.

Finally what I found that iterating over the `fj.event` with `fj.eventIndex` is slower than -
```
data = fj.event[fj.eventIndex:]
i = 0
// iterating from i until len(data)
```

I am not sure why iterating from the middle to from the start is slower, I was thinking about trying to check it with compiler explorer and see what's the difference .. I'll try that but I am not an assembly expert.

flatten_json.go
yosiat: Will add!

flatten_json.go
timbray: Huh... agree, makes no sense to me. `data` is now a new `[]byte` against the same backing memory, maybe Go has some magic about indexing into `[]byte` when the index is small?  Anyhow, if this is repeatable, maybe we should have a fancier version of `step()` and `ch()` that rebases the slice and then all the parts of `flattenJSON` get the benefit?

Also, unless the improvement is dramatic I'm a little reluctant to have one piece of the big complicated function that arbitrarily has its own process for traversing the buffer.

yosiat: > Anyhow, if this is repeatable, maybe we should have a fancier version of step() and ch() that rebases the slice and then all the parts of flattenJSON get the benefit?

I was thinking about giving an option to have an iterator over the event that will hand you the current character and it might does what you are saying, but didn't want to introduce big change like that in this PR.

> Also, unless the improvement is dramatic I'm a little reluctant to have one piece of the big complicated function that arbitrarily has its own process for traversing the buffer.

The improvement was dramatic, I'll post benchmarks differences tomorrow and lets get a decision based on that.

yosiat:
> I'm not a fanatic about coverage percent but when it goes down that always makes me nervous. I looked at the actual coverage red/green and didn't see anything uncovered that looked very worrying, but please have a glance at the coverage and see if there's anything that surprises you.

This is actually interesting! I checked the coverage and increased coverage actually when I ran only `TestFJ` tests from 24%~ to 30%~.

I'll give it a look, it's important that we will have a (almost) perfect coverage for parsing JSON.


I'll push commits to address the changes you are raising and before merging I'll do some rebases with squashes (so it will go with the commit conventions, and will be concentrated together).

flatten_json.go
timbray: > The improvement was dramatic, I'll post benchmarks differences tomorrow and lets get a decision based on that.

Wow, given that ch() and step() are inlined, that is **really** weird (also they didn’t show up in profiling output).  Anyhow, there's a lesson hiding in here somewhere and once we figure it out we should give all of flattener the benefit. I agree, probably another PR for that.

embano1:
Regarding benchmarks: yeah it's tricky, but we could take a look at `https://github.com/benchmark-action/github-action-benchmark` and see if it provides value for us?

flatten_json.go
yosiat: Update on benchmarks, I rewrote `skipStringValue` and `skipBlock` with `step` & `ch` -
```
benchmark                                                old ns/op     new ns/op     delta
Benchmark_JsonFlattener_ContextFields-10                 9062          12211         +34.75%
Benchmark_JsonFlattener_MiddleNestedField-10             10730         13619         +26.92%
Benchmark_JsonFlattener_LastField-10                     9570          12598         +31.64%
Benchmark_JsonFlattner_Evaluate_ContextFields-10         9487          12607         +32.89%
Benchmark_JsonFlattner_Evaluate_MiddleNestedField-10     11122         14021         +26.07%
Benchmark_JsonFlattner_Evaluate_LastField-10             9873          12893         +30.59%

benchmark                                                old allocs     new allocs     delta
Benchmark_JsonFlattener_ContextFields-10                 3              3              +0.00%
Benchmark_JsonFlattener_MiddleNestedField-10             4              4              +0.00%
Benchmark_JsonFlattener_LastField-10                     2              2              +0.00%
Benchmark_JsonFlattner_Evaluate_ContextFields-10         13             13             +0.00%
Benchmark_JsonFlattner_Evaluate_MiddleNestedField-10     13             13             +0.00%
Benchmark_JsonFlattner_Evaluate_LastField-10             9              9              +0.00%

benchmark                                                old bytes     new bytes     delta
Benchmark_JsonFlattener_ContextFields-10                 48            48            +0.00%
Benchmark_JsonFlattener_MiddleNestedField-10             64            64            +0.00%
Benchmark_JsonFlattener_LastField-10                     32            32            +0.00%
Benchmark_JsonFlattner_Evaluate_ContextFields-10         480           480           +0.00%
Benchmark_JsonFlattner_Evaluate_MiddleNestedField-10     488           488           +0.00%
Benchmark_JsonFlattner_Evaluate_LastField-10             312           312           +0.00%
```

I tried to write a representative benchmark - https://gist.github.com/yosiat/930426f4baea92b5f188203c6acb716b and it doesn't makes sense to me, I checked with https://godbolt.org/ and it looks that `step` instructions are the same as `for` code in it's relevant parts 🤔

yosiat:
> Regarding benchmarks: yeah it's tricky, but we could take a look at [benchmark-action/github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) and see if it provides value for us?

This looks really awesome!! Will be interesting to use it here, and get alerts on PRs when we have a regression. I'll do it in PR immediately after this PR is merged, and let's see what we think about the results?

yosiat:
@timbray the PR is ready for another round of review.

Regarding code coverage, the comment says there is a decrease, but when I open https://app.codecov.io/gh/timbray/quamina/pull/119 there is an increase of `0.17%` and for `flatten_json.go` the increase is `0.42%`.

About the commits, once you approve I'll merge, but before merging I'll squash the commits (manually with interactive rebase) and will do force-push.

timbray:
When I look at the codecov report, it says the coverage is up: https://app.codecov.io/github/timbray/quamina/commit/76952717c23cb3f1cc04d1f9f9a61dd23876d876

OK, will review today.

flatten_json.go
```diff
@@ -475,6 +489,65 @@ func (fj *flattenJSON) readLiteral(literal []byte) ([]byte, error) {
 	return literal, nil
 }
 
+func (fj *flattenJSON) skipBlock(openSymbol byte, closeSymbol byte) error {
```
timbray, line 492: So skipBlock sort of works, but I think its behavior is a little troubling.  Try this unit test:

```go
func TestSkipBlock(t *testing.T) {
	blocks := []string{
		`[ { foo ] ]`,
		`{ "foo": }, "bar": 1 }"`,
	}

	for _, block := range blocks {
		f := newJSONFlattener()
		fj := f.(*flattenJSON)
		fj.event = []byte(block)
		fj.eventIndex = 0
		var err error
		if block[0] == '[' {
			err = fj.skipBlock('[', ']')
		} else if block[0] == '{' {
			err = fj.skipBlock('{', '}')
		}
		if err == nil {
			fmt.Printf("for %s block=%s\n", block, string(fj.event[fj.eventIndex:]))
		} else {
			fmt.Printf("for %s err=%s\n", block, err.Error())
		}
	}
}
```

Is this a problem?

yosiat: `skipBlock` comes with a trade-off, that we won't validate the JSON during parsing.. so if we get a block (which we are not interesting about for matching) and contains invalid data we won't err.

To me, it makes sense, it would be problematic if it's a block that we are interesting in and it's not valid.

yosiat: Another perspective is about early skipping that we want to implement next..
consider: event which is `{ "context": ..., "payload": ... }` and we have patterns only on context fields, in that case once we fetch them all we won't read payload.

Now in that case, payload can be invalid json, for example: `{ "context": { "id": 1 }, "payload": { "a": }`.

timbray: OK, I'll be honest, the notion that we would fail to detect broken JSON was sort of shocking to me, and remember that, because of [Hyrum's Law](https://www.hyrumslaw.com), this becomes part of our contract with the world.  So:

We need a change to README.md describing this behavior.  I just checked and the discussion of error-checking in `MatchesForJSONEvent()` is already weak, so I'll give myself an action item to take care of that.

I'll approve the PR but I'd like to request that you do a little investigation and figure out why `ch()` and `step()` seem to be slowing things down, i.e. why the code in `skipBlock` looks different from the rest of `flatten_json.go`.  OK?  If you can't I will, but at this moment you have all this swapped into your head.

timbray (approved):

flatten_json.go
yosiat: I did an investigation, It would be nice if you can enable "Discussions" feature so we can talk about it over there.

I ran some benchmarks, playing around and published the results and the code here: https://github.com/yosiat/quotes-count 

The reason that `step` & `ch` are slower is essentially because they are on the heap, while the other example is on stack. Because of obvious reasons we can move `flattenJson` to be on stack.

The interesting stuff I found is:
* `_QuotesCount_Heap_For_EventIndex` is doing `for eventIndex < len(event)` and incrementing `eventIndex`, on M1 is shows performance benefit while on Intel it doesn't.
* `_QuotesCount_Heap_Next ` - Instead of doing `for` without conditions, I added a `next` method which just checks the eventIndex - on M1 it's faster, but on Intel it's slower - so we can't use this trick.

I checked other json parsers on how they managing their state, and found that it splits:
* Most of them is doing the same as `flattenJson` - referencing their state with heap pointers.
* Others have each method accept the rest of payload to process and return an offset, so for example, reading an array will be `offset, value, err := fj.readArray(fj.event[i:])` and then upon success do `i += offset` , so in both methods the index is on the stack, same as the event.

I can do POC of the second option to see how much it improves the performance, but it will be have worse DX and be more error prone. so I am not sure if the cost (performance vs DX) will be worth it.

What do you think?

timbray: OK, Discussions now enabled and I launched one. I have some questions about your most recent - want to replicate this over in the discussion and I'll follow-up there?

yosiat: Cool! replicated to: https://github.com/timbray/quamina/discussions/122 with some context added for new folks.

PR #124, opened 2022-10-11
Segments Tree Tracking

yosiat:
## Context

Prior discussion is here: https://github.com/timbray/quamina/issues/113

Currently the flattener and more specifically "NameTracker" is not hierarchy aware, to explain that let's examine this case:
Pattern: `{ "context": { "user_id": [1] } }`
Event: `{ "context": { "user_id": 1 }, "payload": { /* lots of fields, large payload */ }`

In this case the flattener will read context and then user_id, all the other values (strings, objects and arrays) will be skipped, but still will get traversed. By making the "NameTracker" aware of the hierarchy, we can:
* Know that once we read "user_id" we can stop consuming the event and exit
* Cache the paths - so `pathForChild` is not needed and we reduce allocations.

## Hierarchy

To add Hierarchy to `NameTracker`, I introduced `SegmentsTreeTracker` interface and `segmentsTree` implementation, which is a simple tree that represents the living patterns, where:
* Node is an object / array we need to traverse into
* Field is leaf value - string/number/array we need to pluck as a Quamina's Field.

The current implementation and idea is to traverse the JSON together with `SegmentsTreeTracker` (you can think about it like a mirror), once we traverse them together we know:
* If the current field in his hierarchy (object/array) is needed - simple lookup on `fields`
* Get the path of a field - simple lookup, since `fields` is map from string (member name) to it's path (`[]byte`)
* Now if we can stop process the object/array - once we traverse all current nodes (objects/arrays) and fields - by decrementing a counter

## Benchmarks

```
name                                         old time/op    new time/op    delta
CityLots-10                                    4.42µs ± 0%    4.28µs ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattener_ContextFields-10                9.06µs ± 0%    0.21µs ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattener_MiddleNestedField-10            10.9µs ± 0%     1.8µs ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattener_LastField-10                    9.65µs ± 0%    9.68µs ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_ContextFields-10        9.42µs ± 0%    0.50µs ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_MiddleNestedField-10    11.0µs ± 0%     2.0µs ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_LastField-10            9.81µs ± 0%    9.87µs ± 0%   ~     (p=1.000 n=1+1)

name                                         old alloc/op   new alloc/op   delta
CityLots-10                                      985B ± 0%      832B ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattener_ContextFields-10                 48.0B ± 0%      0.0B        ~     (p=1.000 n=1+1)
_JsonFlattener_MiddleNestedField-10             64.0B ± 0%      0.0B        ~     (p=1.000 n=1+1)
_JsonFlattener_LastField-10                     32.0B ± 0%      0.0B        ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_ContextFields-10          232B ± 0%      184B ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_MiddleNestedField-10      240B ± 0%       48B ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_LastField-10             80.0B ± 0%     48.0B ± 0%   ~     (p=1.000 n=1+1)

name                                         old allocs/op  new allocs/op  delta
CityLots-10                                      38.0 ± 0%      31.0 ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattener_ContextFields-10                  3.00 ± 0%      0.00        ~     (p=1.000 n=1+1)
_JsonFlattener_MiddleNestedField-10              4.00 ± 0%      0.00        ~     (p=1.000 n=1+1)
_JsonFlattener_LastField-10                      2.00 ± 0%      0.00        ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_ContextFields-10          9.00 ± 0%      6.00 ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_MiddleNestedField-10      9.00 ± 0%      3.00 ± 0%   ~     (p=1.000 n=1+1)
_JsonFlattner_Evaluate_LastField-10              5.00 ± 0%      3.00 ± 0%   ~     (p=1.000 n=1+1)
```

We can see the expected results:
* `CityLots` and `LastField` have improved only by allocations (for paths, no `pathForChild` calls), but speed isn't improved since we need to consume whole event here.
* `Context` & `MiddleNestedField` have improved by a lot since we can do early stop and we don't allocate for paths.

## What's next & Status

This PR is not ready to be merged, I am opening it to get an early feedback (given all tests passing, benchmark results) about the interface and implementation, once I get approval on that I can continue to:

- [x]  Adding tests: lots of tests are missing here - for paths indexing, objects as fields, etc..
- [ ]  Investigate degradation of `LastField`
- [x]  Improve the early stop case - currently it's funky with returning error and skipping object till the end, tests are needed here.
- [x]  Objects as fields - Currently this case is not covered, but assuming we have "exists: true" pattern on object, it shouldn't pass (didn't checked if we have test for it)
- [x]  Remove `NameTracker` and `namesUsed`

### Concern about skipping

We talked about this in the previous PR (https://github.com/timbray/quamina/pull/119), that we skip JSONs and now a bit more aggressive, and we can accept invalid JSON.

I have an idea on how to solve it - since there is a tradeoff about strictness and performance, I suggest let the user decide what he prefers with an option to flattener - if "strict mode" is enabled we won't do any skipping and no early exists.

If this is something we will to consider, I'll create separate issue and PR for it.


yosiat:
@timbray PR is not ready for (initial) review, I want to rebase it a bit and clean the commits, I'll ping you (in a few minutes) when it will be ready for your review.

embano1:
This is some great insights! On a related note, I posted a similar question/thought in the [`ruler`](https://github.com/aws/event-ruler/issues/46) project, which Tim was also involved. Not sure your findings translate 1:1, but this confirms my assumption of speeding up matching by the `context` (or `metadata`) use case, e.g. `CloudEvents` envelop.

timbray:
Sorry Yosi, I'm maxed out on another deliverable, I'll get to this PR by end-of-day or latest tomorrow morning.

yosiat:
@timbray No rush! Review when you have time, I will currently try to work on the other PRs I mentioned I have in-line.

arrays_test.go
```diff
@@ -70,30 +70,29 @@ const bands = `{
 }`
 
 func TestArrayCorrectness(t *testing.T) {
-	// only pattern3 should match
-	pattern1 := `{"bands": { "members": { "given": [ "Mick" ], "surname": [ "Strummer" ] } } }`
-	pattern2 := `{"bands": { "members": { "given": [ "Wata" ], "role": [ "drums" ] } } }`
-	pattern3 := `{"bands": { "members": { "given": [ "Wata" ], "role": [ "guitar" ] } } }`
+	// only wataGuiterPattern should match
```
timbray, line 73: a bit curious why the renaming. More readable for sure!

citylots_bench_test.go
```diff
@@ -0,0 +1,60 @@
+package quamina
```
timbray, line 1: Assuming this is just extraction to a new file… any logical changes to the benchmark?

core_matcher.go
```diff
@@ -70,14 +72,22 @@ func (m *coreMatcher) addPattern(x X, patternJSON string) error {
 	freshStart := &coreStart{}
 	freshStart.namesUsed = make(map[string]bool)
 	current := m.start()
+	freshStart.pathsIndex = current.pathsIndex.copy() // NOTE: not sure if this is thread-safe
```
timbray, line 75: Hmm, looks thread-safe to me but there's any doubt should do concurrency fuzz-testing.

flatten_json.go
```diff
@@ -208,42 +227,56 @@ func (fj *flattenJSON) readObject(pathName []byte) error {
...
-				}
 
 				if fj.skipping > 0 || !memberIsUsed {
 					err = fj.skipBlock('[', ']')
 				} else {
-					err = fj.readArray(arrayPath)
+					arrayPathNode, exists := pathNode.Get(memberName)
+					if !exists {
+						// Arrays are interesting, they can be field or node.
+						// Given this case:
```
timbray, line 240: Really like this style of comment

flatten_json.go
```diff
@@ -208,42 +227,56 @@ func (fj *flattenJSON) readObject(pathName []byte) error {
...
 					err = fj.skipBlock('{', '}')
 				} else {
-					err = fj.readObject(objectPath)
+					objectPathNode, exists := pathNode.Get(memberName)
+					if !exists {
+						// Interesting case.
+						// Looks like it's an object, which not registered as object (node) in our paths index.
+						// Which means someone is doing matching on object field as primitive field, for example:
+						//  pattern: `{ "a": [1] }`, event: `{ "a": { "b": 1 } }` for example.
+						return fj.error(fmt.Sprintf("[%s] doesn't exists in path index, but we still got here [%s].\n", string(memberName), pathNode.String()))
```
timbray, line 268: Haven't fully studied yet, thinking out loud: Is this a situation that can arise, i.e. if you add that pattern and match that event, would you get to this `fj.error`?  If so, it shouldn't be an error, it should be just a no-match return?  If you can't get to this error, should the code even exist?

flatten_json.go
```diff
@@ -208,42 +227,56 @@ func (fj *flattenJSON) readObject(pathName []byte) error {
...
 				}
-				var objectPath []byte
-				if fj.skipping == 0 {
-					objectPath = pathForChild(pathName, memberName)
-				}
 				if fj.skipping > 0 || !memberIsUsed {
 					err = fj.skipBlock('{', '}')
 				} else {
-					err = fj.readObject(objectPath)
+					objectPathNode, exists := pathNode.Get(memberName)
```
timbray, line 262: would `ok` be a bit more idiomatic than `exists`? (Not a rhetorical question, maybe lots of people use things other than `ok` and I just haven't seen it.)

name_tracker.go
```diff
@@ -6,14 +6,47 @@ package quamina
...
-//  "a\nb" 1
-//  "a\nc" 2
+//
+//	"a\nb" 1
+//	"a\nc" 2
+//
 // Note that the values stored in NameTracker are not the full paths "a\nb" and "a\nc", but the path segments
 // "a", "b", and "c".  This reduces the work the Flattener has to do - whenever it processes an object, it need
 // only look up the member names in the NameTracker. A little thought reveals that this can produce false
 // positives and some potential wasted work, but is a good trade-off.
```
timbray, line 20: Maybe a remark here about the new hierarchical tracker so people understand that NameTracker is not the only optimization in play?

name_tracker.go
```diff
@@ -6,14 +6,47 @@ package quamina
...
 type NameTracker interface {
 	IsNameUsed(label []byte) bool
+
+	// This should be actually part of matcher interface and not NameTracker.
```
timbray, line 24: Why not put it in the matcher interface then? We have not done a 1.0 release, we can change APIs if we want to.

name_tracker.go
```diff
@@ -6,14 +6,47 @@ package quamina
...
+	GetHierarchalNameTracker() HierarchalNameTracker
+}
+
+type HierarchalNameTracker interface {
```
timbray, line 28: I think it would be nice to have a long-ish comment here introducing what this type exists to do and why it has this design.

I'm also wondering about the `Name` in the type name. JSON brings the concept of "member name" and then Quamina has the concept of "Path" and then "segment" (which is actually just a member name). Is it possible to give this type a name based on these existing comments?  Or perhaps a name based on what useful thing it actually **does** e.g. "pathSkipper" or some such?

name_tracker.go
```diff
@@ -6,14 +6,47 @@ package quamina
...
+	// Get returns another level of the hierarchy, referred as "Node"
+	// If a node is returned we will need to traverse into (in JSON/CBOR/ProtoBuf/etc..)
+	Get(member []byte) (HierarchalNameTracker, bool)
+
+	// IsRoot - are we root node?
+	// NOTE: need for early exit, can be solved differently maybe.
+	IsRoot() bool
+
+	// Is the current property needs to be plucked as a Quamina's Field.
```
timbray, line 37: Not sure what "property" means… field? Probably "processed" rather than "plucked"?

paths.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+const PathSeparator = "\n"
+
+type pathsIndex struct {
+	root bool
+
+	nodes map[string]*pathsIndex
+
+	// fields map from it's name to it's full path (as specific in quamina.Field)
```
timbray, line 15: s/it's/its/ because "it's" == "it is". Like we say "he's" for "he is" but not "hi's".

paths.go
```diff
@@ -0,0 +1,128 @@
...
+	// will be present only on the leafs
+	fields map[string][]byte
+}
+
+func newPathsIndex() *pathsIndex {
```
timbray, line 20: maybe `newRootIndex` or some such, since it's not general-purpose but only used for root node?

paths.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (p *pathsIndex) Get(name []byte) (HierarchalNameTracker, bool) {
+	n, ok := p.nodes[string(name)]
+	return n, ok
+}
+
+func (p *pathsIndex) getOrCreate(name string) *pathsIndex {
+	if _, ok := p.nodes[name]; !ok {
```
timbray, line 56: For consistent style, please break into 2 lines. 
```go
   _, ok := p.nodes[name]
  if !ok { ...
```
(I hate trailing `.   }\n. }\n }`)

(recurs)

paths.go
```diff
@@ -0,0 +1,128 @@
...
+	if _, ok := p.fields[name]; !ok {
+		p.fields[name] = path
+	}
+}
+
+func (p *pathsIndex) IsRoot() bool {
+	return p.root
+}
+
+func (p *pathsIndex) IsNameUsed(name []byte) bool {
```
timbray, line 77: for clarity, maybe argument should be called either `segment` or `path`?

paths.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (p *pathsIndex) NodesCount() int {
+	return len(p.nodes)
+}
+
+func (p *pathsIndex) FieldsCount() int {
+	return len(p.fields)
+}
+
+func (p *pathsIndex) String() string {
```
timbray, line 98: is this for debugging/testing or actually called at runtime or in error-handling? If just for debug/test, please have a comment saying so.  BTW I'm totally in favor of having `String()` func for any reasonably complex type to facilitate debugging.

paths.go
```diff
@@ -0,0 +1,128 @@
+package quamina
+
+import (
+	"fmt"
+	"strings"
+)
+
+const PathSeparator = "\n"
+
+type pathsIndex struct {
```
timbray, line 10: If this implements HierarchicalNameTracker then I think it's polite to have a comment that says so. I'm mostly not 100% a fan of Go's silent interface implementation.

timbray (commented):
I think I agree with this but I'm not 100% sure I understand.  Most of my comments here are minor but one is important: Could you put a long-ish explanatory comment on type HierarchicalNameTracker (and also maybe on the implementation in paths.go) explaining at a high level what this does. Lots of the central type implementations in Quamina have such a comment.  

I think this is especially important because we anticipate in future that people will also want to write Flatteners for Avro and Protobufs and CBOR etc etc etc and I guess they will have to duplicate some of this work?  In that spirit, is the code in paths.go specific to JSON?

After you've written this, you may find you want to rename some of the types and methods… what is `Get()` getting? What is the `Name` in HierarchicalNameTracker? etc.

timbray:
BTW, I think the verb "pluck" is maybe not the best choice. In English it has two meanings which are opposite from each other. (English has way too many words like this.)

1. I plucked the best apple from the tree
2. I plucked the feathers from the chicken (i.e. removed them to throw away)

So the message is a bit confusing, do you pluck to select or pluck to discard?

yosiat:
Thanks for the review!

> I think I agree with this but I'm not 100% sure I understand. Most of my comments here are minor but one is important: Could you put a long-ish explanatory comment on type HierarchicalNameTracker (and also maybe on the implementation in paths.go) explaining at a high level what this does. Lots of the central type implementations in Quamina have such a comment.

I will add such comment, but want to see if you accept the interface and the general idea.

>  I guess they will have to duplicate some of this work? In that spirit, is the code in paths.go specific to JSON?

The idea behind this interface and `paths.go` is to have a generic implementation that can be used for JSON and other future Flatteners. All of them are flattening hierarchal data, The difference might comes that ones won't need to traverse the hierarchy and he will a fast way to get value via path pointers, and this tree data structure might not be 100% efficient for them.

I think that we will need to challenge the Flattener with writing another one for data format which is different than JSON, I want to do it, but it's a good amount of work and time that I don't have right now, but I am planning to do it. Not sure when though.

> After you've written this, you may find you want to rename some of the types and methods… what is Get() getting? What is the Name in HierarchicalNameTracker? etc.

You wrote about this here and in the comments and I agree with you, I didn't maintained a consistent terminology - sometime it's a member, sometimes it's a property and sometimes it's a name.

Now that I understand (from you comments) the terminology in Quamina I'll make a names across the code in this PR.

Just one question for me to you, now that I understand that we have "Path" and it's parts are called "Segment", why you called it `NameTracker` and not `SegmentTracker` ?

> So the message is a bit confusing, do you pluck to select or pluck to discard?

I meant Pluck as Select and not discard :)

I used pluck because other languages and libraries have a method called `pluck` on object/hash which allows to select specific fields, like rails for example: https://apidock.com/rails/ActiveRecord/Calculations/pluck and in Ruby actually it's called select.

I'll rename pluck to select.

arrays_test.go
yosiat: I had a bug which failed this test and it was hard for me to debug it because of the names `pattern1` / `pattern2` etc so I did those changes.

BTW, I think Quamina testing will benefit by adding good assertions library (like https://github.com/stretchr/testify)

citylots_bench_test.go
yosiat: It's the same benchmark code, the main difference (and this why I kept the original) is that I couldn't go benchmarks to emit events per second, which is an important metric.

Anyway, I am going to remove this benchmark from this PR and add it (with CI as discussed in the past) here - https://github.com/timbray/quamina/pull/126 (not yet ready for review)

core_matcher.go
yosiat: It looks thread-safe to me as well, but I wasn't sure what happens if during the `copy` another go-routine changes the `nodes` / `fields` (important to note we have only additions, not deletions). The problem with that in theory that we might loose changes, since we are working with specific snapshot.

I'll do concurrency fuzz-testing, never did one, but will look at `concurrency_tests` as a guidance.

flatten_json.go
yosiat: Ignore the error returning here - I did this for debug purposes and in the final PR it should be removed.

I wasn't sure (and didn't checked) if we support select objects as `quamina.Field`, the use-case I have in mind is doing `exists: true` on object (this is something I plan doing).

An example for that will be, given this event: `{ "name": "yosi", "changes": { "name": ["yosy", "yosi"] } }` I would like to get match only when `changes` exists in the object (since some events might not contains `changes` json property).

If we want to support in the future/currently support it, it means that this code should get same handling I did for arrays: https://github.com/timbray/quamina/pull/124/files#diff-745034a9cc805c7b30a993232b1a6d53132a7e2099d5b72641b40ddfa21e3916R239

flatten_json.go
yosiat: `exists` is a problem of mine :) I am used for using `exists` as variable name since it makes more sense to me, but I know the go community went for `ok` - so I'll align this code and use `ok` (even though it's not natural for me, it's just better to align with conventions).

name_tracker.go
yosiat: Will move 👍

name_tracker.go
yosiat: Talked about it here - https://github.com/timbray/quamina/pull/124#issuecomment-1277377837

About a name, what it's actually doing in practice is "allowing to efficiently select an hierarchical data", so "pathSkipper" don't think it's a good name, what about `SegmentsTreeTracker` ? (I'm really bad a this =X)

name_tracker.go
yosiat: It should be actually (after the learnings today on Quamin's internal terminology) -
```
// Do we need to select the given segment from event as Quamina's Field.
IsSegmentUsed(segment []byte) bool
```

paths.go
yosiat: Actually it's being used in `copy` on all nodes both root and non-root, I'll refactor it so I'll have two methods one for root and another for nodes.

paths.go
yosiat: It's for debugging, will add a comment for that 👍

paths.go
yosiat: > I'm mostly not 100% a fan of Go's silent interface implementation.

I always had a conflicted feelings coming from C# to Go about this, and will add a comment to help with those feelings :)

yosiat:
@timbray 

Thanks for this amazing review, learned a lot, I'll add most of the comments, there are some that waiting for three things to be clarified/answered (and once those are clarified, I'll fix all comments):
1. `why you called it NameTracker and not SegmentTracker`
2. Do we want to keep `NameTracker` and `namesUsed` discussed here - https://github.com/timbray/quamina/pull/124#discussion_r994467358
3. Allowing to select objects as `quamina.Field` - https://github.com/timbray/quamina/pull/124#discussion_r994461942

Thanks 🙏

arrays_test.go
timbray: I'm used to that kind of thing from my years in Java-land. Hmm, I don't find writing tests of the form
```go
if x != nil {
  t.Error("whatever")
}
```

But I guess the advantage is that with the assertions and so on, the tests are more readable? That seems like a good argument.

core_matcher.go
timbray: But, the event body, and then the nodes and fields, are only visible to the thread that's running `matchesForSortedFields`, right?

flatten_json.go
timbray: Right, at the moment `"exists":true` only works for leaf nodes, because of the way flattening works. I would 100% agree in principle with fixing this but it would be a nontrivial change to the state-machine model.  I suggest that we shouldn't hold up this PR for that enhancement but let's record an issue so we don't lose the idea.

timbray:
I agree, NameTracker certainly isn't the best name. Happy to accept something better. 

As I said, my impression is that making `"exists":true` work is nontrivial, but maybe you found a good way to do it and I didn't notice?  I don't want to hold up this nice optimization to wait for that if it's difficult.

name_tracker.go
yosiat: I remembered now why I did it like this! It means that `Flatten` will need to accept a `Matcher` interface instead.

Maybe a composed interface -
```
type Tracker interface {
  NameTracker
  HierarchalNameTracker
}
```

Or - 
```
type Tracker interface {
  GetNameTracker()
  GetHierarchalNameTracker()
}
```

?

name_tracker.go
yosiat: renamed

paths.go
yosiat: Added comments.

paths.go
yosiat: Added a comment.

paths.go
yosiat: Rename all names/properties to segments.

paths.go
yosiat: Change the occurrences in this file.

paths.go
yosiat: Refactored this method to two methods.

paths.go
yosiat: Thanks, changed!

name_tracker.go
yosiat: Added a comment an renamed - interface to `SegmentsTreeTracker` and `pathsIndex` to `segmentsTree`, if you have a better names I'll replace it :)

flatten_json.go
yosiat: Renamed to "ok"

name_tracker.go
yosiat: @timbray `do we want to keep use of NameTracker and namesUsed ?` what's your opinion on that?

flatten_json.go
yosiat: I checked again for what I specifically need, and I was wrong:
```
{ "name": "yosi", "changes": { "name": ["yosy", "yosi"] } }
```

I would need `exists: true` on `changes.name` and not on `changes`.

Now the question is:
* Do we still want an issue for allowing `exists: true` on object nodes?
* How should I handle this case, because currently it's an error, I will try to reproduce in main and see what's the behaviour.

arrays_test.go
yosiat: The benefit of assertions library is increasing readability of tests, in addition to better error messages when tests fail (for example: comparing two arrays, it will output both arrays to the screen with delta).

I wrote this as a note to myself, to check an assertion library for Quamina and purpose it, I'll make some examples so we can have an elaborate discussion on a draft PR.

flatten_json.go
yosiat: Fixed the login here, and added tests for this to match the main branch:
https://github.com/timbray/quamina/pull/124/commits/1080360922fe4a84208fb1c5070a9f35eeee7b6a

Once we want to add support for "exists" on objects we can revisit this.

core_matcher.go
yosiat: Yep you are right, After thinking about this again, I don't see any issue with thread-safety here. removing this comment.

timbray:
This PR ready for another review?

yosiat:
@timbray yes, I fixed most of the comments, added tests & fixed some.

The major thing we didn't answered about, is "NameTracker" & "namesUsed", do we still want to keep it? it's affecting both:
* Code coverage - since "IsNameUsed" is not used anymore
* The interface that flattener accepts, we had a discussion about this.

I am not sure where you are standing about, so we can do an elaborated discussion on that matter, but from my side:
* I don't believe in two solutions both "SegmentsTree" and "NameTracker", it will be bad for us to maintain and external flatteners to work with.
* I proved that "SegmentsTree" is a good solution for JSON flattening, but didn't prove it's good in general for flattening, I think it can be, but to actually validate it an additional flattener will need to be written (protobuf/cbor).

timbray:
I 109% agree on refactoring and improving the names. Do you want to do that before I look again?

yosiat:
@timbray I think you missed me If I read your comment correctly.

I did all the naming changes, the only thing left to decide is *can we remove NameTracker and replace it with SegmentsTree* ?

timbray:
Oh, absolutely, yes, please replace.  I'll throw another review on it after I finish writing the extremely boring doc I'm working on (maybe not till tomorrow).

yosiat:
Cool! I'll do the replacements tomorrow, it will take some time since I wrote all tests to be compat to "NameTracker", will write here once I am finished.

yosiat:
@timbray The pr is ready for your re-review.

timbray:
> @timbray The pr is ready for your re-review.

Excellent. Sorry, won't get there for a day or so.  Is there a good place to start, i.e. a big comment explaining the whole segmentsTree stuff?

yosiat:
@timbray to review this PR I suggest:
1. Read the pull request body first
2. Read the interface & it's comment: segments_tree_tracker.go
3. Read the implementation - segments_tree.go (and it's tests to understand it's usage a bit)
4. Then it's first usage which is core_matcher in pattern addition
5. Lastly json flattener

segments_tree.go
```diff
@@ -0,0 +1,141 @@
...
+		// It's a direct field.
+		p.fields[path] = []byte(path)
+		return
+	}
+
+	var node *segmentsTree
+
+	node = p.getOrCreate(parts[0])
+
+	for i, part := range parts[1:] {
```
timbray, line 51: This logic is baffling me.  A comment might help?

segments_tree.go
```diff
@@ -0,0 +1,141 @@
+package quamina
+
+import (
+	"fmt"
+	"strings"
+)
+
+const SegmentSeparator = "\n"
+
+type segmentsTree struct {
```
timbray, line 10: So far I haven't figured out how this works at all.  I think for a type like this it's useful and conventional to have a comment block explaining the fields, as most of the other significant Quamina types do.

segments_tree.go
```diff
@@ -0,0 +1,141 @@
...
+	return n, ok
+}
+
+// IsRoot implements SegmentsTreeTracker
+func (p *segmentsTree) IsRoot() bool {
+	return p.root
+}
+
+// IsSegmentUsed implements SegmentsTreeTracker
+func (p *segmentsTree) IsSegmentUsed(segment []byte) bool {
```
timbray, line 88: Another place that's hard to understand, could use a little commenting.

timbray (commented):
I need to take a break and look at some other stuff.  I haven't seen anything wrong with the code but I still haven't figured out how segmentsTree works.  I've got your branch checked out and writing some tests to build the tree and using the debugger to poke around inside, so I'll figure it out eventually. I think what I need is to work a straightforward example and see how we get from `segmentsTree.add()` to `pathNode.isSegmentUsed()`. I'm sure I'll get it then. Will review more later.

segments_tree.go
yosiat: Re-wrote the implementation and added a comment

segments_tree.go
yosiat: added comment

segments_tree.go
yosiat: I wrote such a comment on the interface it self `segments_tree_tracker.go`, adding a comment on segmentsTree will be the same comment from my side.

But I added some comments around the code - what is fields, nodes and for other methods with examples, hopefully it's clearer now.

segments_tree.go
```diff
@@ -0,0 +1,141 @@
...
+func newSegmentsIndexNode(root bool) *segmentsTree {
+	return &segmentsTree{
+		root:   root,
+		nodes:  make(map[string]*segmentsTree),
+		fields: make(map[string][]byte),
+	}
+}
+
+func (p *segmentsTree) add(path string) {
+	parts := strings.Split(path, SegmentSeparator)
```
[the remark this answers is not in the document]
yosiat: renamed :)

segments_tree.go
timbray: Ah, I had looked at the interface definition but hadn't really absorbed it, but now after looking at the code, this is making more sense. BTW, after we merge, I will probably do a PR to improve spelling and English language style (I hope that's OK) but it's not worth holding this up.

segments_tree.go
```diff
@@ -10,13 +10,20 @@ const SegmentSeparator = "\n"
...
 	root bool
 
+	// "nodes" stores a map from a segment to it's childrens
+	// in an hierarchial data like JSON, a node can be Object or Array.
+	// for example, in this path "context\nuser\nid", both "context" and "user" will be nodes.
 	nodes map[string]*segmentsTree
 
-	// fields map segment to full path
-	// will be present only on the leafs
+	// "fields" is storing leaf fields here mapped to their full paths
```
timbray, line 18: I suggest we find a different name than "fields" because `Field` is already a well-known widely-used Quamina concept.  I notice your comment says "…is storing leaf nodes" so maybe "leaves" would be a better name?  Or something else.  

Hmm, thinking out loud, the leaf name is the key and the value is (in Quamina terms) `Field.Path`. So maybe this could be called `leafPaths` or something?

segments_tree.go
```diff
@@ -0,0 +1,155 @@
...
+	return len(p.nodes)
+}
+
+// FieldsCount implements SegmentsTreeTracker
+func (p *segmentsTree) FieldsCount() int {
+	return len(p.fields)
+}
+
+// String used for debugging purposes
+func (p *segmentsTree) String() string {
```
timbray, line 127: BTW I very much like having `String` for complex types.  I have been creating them for debugging and then throwing them away, but looking at this, I think I should leave them in.

segments_tree.go
```diff
@@ -0,0 +1,155 @@
...
+// newSegmentsIndexNode initializes a segmentsTree node
+func newSegmentsIndexNode(root bool) *segmentsTree {
+	return &segmentsTree{
+		root:   root,
+		nodes:  make(map[string]*segmentsTree),
+		fields: make(map[string][]byte),
+	}
+}
+
+func (p *segmentsTree) add(path string) {
```
timbray, line 46: Thanks, much clearer now.  I still haven't 100% figured this out, but I think I'm starting to get it. The `[len-2]` was breaking my brain.

timbray (commented):
Just one serious comment, about the name "fields".  I'm getting closer to understanding, I'm nearly 100% sure I like the design but I need to really understand it first.  Quoting Prof Feynman: https://twitter.com/proffeynman/status/1370595934191591430?lang=en

segments_tree.go
yosiat: Yeah, the `len-2` and skipping the first element was too complex.. this method can be re-written BTW, let me know what you think about it:

```go
	// Add the nodes
	for _, segment := range segments[:len(segments)-1] {
		node = node.getOrCreate(segment)
	}
	// Add the leaf
	node.addSegment(segments[len(segments)-1], []byte(path))
```

segments_tree.go
yosiat: Yeah, the name could be better here.. I couldn't figure out which name to call it.

If we are talking about JSON-land it should be named "properties", if we are talking about Quamina's land it should be called "segments" but it's the last segment of path.

When we will call it "leaves" it will signal that this data is stored only on leaf nodes (in our tree), but this data can be actually stored at any level, think about the following paths (using dot as separator):
1. `context.user_id` - will be stored on the `context` node (under root)
2. `context.user.id` - will be stored on the `user` node (under `context`)
3. `user_id` - will be stored on the root node

`leafPaths` is ok, but don't you think it will be confusing since we are storing segments as keys?

timbray:
Going to push exists-false and that'll free up my Quamina time to get your PR landed. But you'll have to rebase a bit, sorry.

yosiat:
@timbray I pushed a rebase :)

BTW, I saw that benchmarks fails on regression in master (https://github.com/timbray/quamina/actions/runs/3430881550/jobs/5718364564), this PR will help with this regression.

core_matcher.go
```diff
@@ -31,15 +31,12 @@ type coreMatcher struct {
 
 // coreFields groups the updateable fields in coreMatcher.
 // state is the start of the automaton.
-// namesUsed is a map of field names that are used in any of the patterns that this automaton encodes. Typically,
+// segmentsTree is a tree of segments that are used in any of the patterns that this automaton encodes.Typically,
 // patterns only consider a subset of the fields in an incoming data object, and there is no reason to consider
 // fields that do not appear in patterns when using the automaton for matching.
-// fakeField is used when the flattener for an event returns no fields, because it could still match if
```
yosiat, line 37: @timbray FYI - I removed the "fakeField" documentation since you removed it from the struct, hope it's ok.

timbray: Thanks.

flatten_json.go
```diff
@@ -15,14 +16,14 @@ import (
...
-	arrayCount int32       // how many arrays we've seen, used in building arrayTrail
-	tracker    NameTracker // knows if a segment is used; if not, no need to process
-	cleanSheet bool        // initially true, don't have to call Reset()
+	event      []byte              // event being processed, treated as immutable
+	eventIndex int                 // current byte index into the event
+	fields     []Field             // the under-construction return value of the Flatten method
+	skipping   int                 // track whether we're within the scope of a segment that isn't used
+	arrayTrail []ArrayPos          // current array-position cookie crumbs
+	arrayCount int32               // how many arrays we've seen, used in building arrayTrail
+	tracker    SegmentsTreeTracker // knows if a segment is used; if not, no need to process
```
timbray, line 25: Not obvious why tracker needs to be in the fj struct.   In flattenJSON you pass the `pathnode` argument.

timbray (commented):
OK, it took me a while with Covid walking through our house, but I got it.  I only have one question then we can merge.

flatten_json.go
yosiat: You are correct, it's not needed.. removed.

timbray (approved):

Issue #115, opened 2022-09-28
bug: exists:false has issues


timbray:
Check out closed PR https://github.com/timbray/quamina/pull/104

Significant redesign of "exists": false required.

PR #141, opened 2022-11-06
bug: Make exists:false work

timbray:
closes: #115

Thanks once again to [kylemcc](https://github.com/kylemcc) for finding this bug and providing the first unit test.

The solution here is clumsy and slow-ish but I think correct.

@yosiat Do we have a conflict with your big PR? There are some changes to field_matcher.go. Just wondering about which order is smartest to merge them in, after reviewing of course.

### DIFF
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -66,6 +66,40 @@ are two Exists Patterns that would match the Events above:
 If a Field in a Pattern contains an Exists Pattern, it
 **MUST NOT** contain any other values.
 
+Exists Patterns currently only work on leaf nodes. That is to 
+say, given this event:
+
+```json
+{ "a": { "b":  1 } }
+```
+
+The following pattern will not match:
+
+```json
+{ "a": [ {"exists": true} ] }
+```
+
+We may be able to change this in future.
+
+The case of empty arrays is interesting. Consider this event:
+
+```json
+{ "a": [] }`
+```
+
+Then `"exists": true` does not match but `"exists": false` does.
+I.e., only the first of the two sample patterns below matches.
+
+```json
+{ "a":  [ { "exists": false } ] }
+```
+```json
+{ "a":  [ { "exists": true } ] }
+```
+This makes sense in the context of the leaf-node semantics; there
+really is no value for the `"a"` field.
+
+
 ### Anything-But Pattern
 
 The Pattern Type of an Anything-But Pattern is
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -84,7 +84,7 @@ We may be able to change this in future.
 The case of empty arrays is interesting. Consider this event:
 
 ```json
-{ "a": [] }`
+{ "a": [] }
 ```
 
 Then `"exists": true` does not match but `"exists": false` does.
--- a/README.md
+++ b/README.md
@@ -131,6 +131,10 @@ one `*` character. The architecture probably allows
 support for a larger subset of regular expressions, 
 eventually.
 
+The `"exists":true` and `"exists":false` patterns 
+have corner cases; details are covered in
+[Patterns in Quamina](PATTERNS.md).
+
 Number matching is weak - the number has to appear 
 exactly the same in the Pattern and the Event. I.e.,
 Quamina doesn't know that 35, 35.000, and 3.5e1 are the
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ have corner cases; details are covered in
 Number matching is weak - the number has to appear 
 exactly the same in the Pattern and the Event. I.e.,
 Quamina doesn't know that 35, 35.000, and 3.5e1 are the
-same number. There's a fix for this in the code which 
+same number. There's a fix   for this in the code which 
 is not yet activated because it causes a 
 significant performance penalty, so the API needs to
 be enhanced to only ask for it when you need it.
--- a/core_matcher.go
+++ b/core_matcher.go
@@ -19,34 +19,48 @@ import (
 )
 
 // coreMatcher uses a finite automaton to implement the matchesForJSONEvent and MatchesForFields functions.
-// state is the start of the automaton
-// namesUsed is a map of field names that are used in any of the patterns that this automaton encodes. Typically,
-// patterns only consider a subset of the fields in an incoming data object, and there is no reason to consider
-// fields that do not appear in patterns when using the automaton for matching
-// the updateable fields are grouped into the coreStart member so they can be updated atomically using atomic.Load()
+// The updateable fields are grouped into the coreFields member so they can be updated atomically using atomic.Load()
 // and atomic.Store(). This is necessary for coreMatcher to be thread-safe.
 type coreMatcher struct {
-	updateable atomic.Value // always holds a *coreStart
+	updateable atomic.Value // always holds a *coreFields
 	lock       sync.Mutex
 }
-type coreStart struct {
-	state                     *fieldMatcher
-	namesUsed                 map[string]bool
-	presumedExistFalseMatches *matchSet
+
+// coreFields groups the updateable fields in coreMatcher.
+// state is the start of the automaton.
+// namesUsed is a map of field names that are used in any of the patterns that this automaton encodes. Typically,
+// patterns only consider a subset of the fields in an incoming data object, and there is no reason to consider
+// fields that do not appear in patterns when using the automaton for matching.
+// fakeField is used when the flattener for an event returns no fields, because it could still match if
+// there were patterns with "exists":false. So in this case we run one fake field through the matcher
+// which will cause it to notice that any "exists":false patterns should match.
+type coreFields struct {
+	state     *fieldMatcher
+	namesUsed map[string]bool
+	fakeField []Field
 }
 
 func newCoreMatcher() *coreMatcher {
+	// because of the way the matcher works, to serve its purpose of ensuring that "exists":false maches
+	// will be detected, the Path has to be lexically greater than any field path that appears in
+	// "exists":false. The value with byteCeiling works because that byte can't actually appear in any
+	// user-supplied path-name because it's not valid in UTF-8
+	fake := Field{
+		Path:       []byte{byte(byteCeiling)},
+		Val:        []byte(""),
+		ArrayTrail: []ArrayPos{{0, 0}},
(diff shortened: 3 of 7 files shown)


core_matcher.go
```diff
@@ -19,34 +19,48 @@ import (
...
-// namesUsed is a map of field names that are used in any of the patterns that this automaton encodes. Typically,
-// patterns only consider a subset of the fields in an incoming data object, and there is no reason to consider
-// fields that do not appear in patterns when using the automaton for matching
-// the updateable fields are grouped into the coreStart member so they can be updated atomically using atomic.Load()
+// The updateable fields are grouped into the coreFields member so they can be updated atomically using atomic.Load()
 // and atomic.Store(). This is necessary for coreMatcher to be thread-safe.
 type coreMatcher struct {
-	updateable atomic.Value // always holds a *coreStart
+	updateable atomic.Value // always holds a *coreFields
 	lock       sync.Mutex
```
embano1, line 26: do we still need this lock then?

core_matcher.go
```diff
@@ -19,34 +19,48 @@ import (
...
+// namesUsed is a map of field names that are used in any of the patterns that this automaton encodes. Typically,
+// patterns only consider a subset of the fields in an incoming data object, and there is no reason to consider
+// fields that do not appear in patterns when using the automaton for matching.
+// fakeField is used when the flattener for an event returns no fields, because it could still match if
+// there were patterns with "exists":false. So in this case we run one fake field through the matcher
+// which will cause it to notice that any "exists":false patterns should match.
+type coreFields struct {
+	state     *fieldMatcher
+	namesUsed map[string]bool
+	fakeField []Field
```
embano1, line 40: From what I can tell how this is used this can be simplified:

- drop this field
- create and use a global variable instead

Wondering if we need this rather "static" object in the struct and instead make it a global variable?

```var fakeField = Field{
 		Path:       []byte{byte(byteCeiling)},
 		Val:        []byte(""),
 		ArrayTrail: []ArrayPos{{0, 0}},
 	}
```

```
	if len(fields) == 0 {
		fields = []Field(fakeField)
	}
```

embano1: Or even simpler (without needing the global variable):

```
	if len(fields) == 0 {
		fields = []Field{<construct inline fake field>}
	}
```

core_matcher.go
timbray: Yeah, but I should put in a comment explaining what's up.  The lock is so that only one thread can update the state machines at a time. But there can be many threads matching events with read-only access in concurrently with the updating thread, and the atomic.Value is so that the updating thread can update the maps and slices while the other threads are reading them.

core_matcher.go
timbray: Yes, that's the right thing. Earlier, I thought that the fakeField would be required in several places, so I stored it in the coreMatcher object, since there's only one coreMatcher. I suppose a global variable would be better but this is even better.

jsmorph (commented):
Is it worth noting in `PATTERNS.md` how empty arrays work with `"exists"`?

1. `"exists":false` with `[]` matches
2. `"exists":true` with `[]` does not match

(I just checked with `aws events test-event-pattern` to be sure.)

The behavior with `null` values seems obvious enough (since `null` _is_ a value).

Maybe not now since this behavior is consistent with EventBridge.. (A specification of some sort would be nice at some point.)

jsmorph:
FWIW, I did sketch the following quickly re some edge cases.

```Go
func TestExistsEdges(t *testing.T) {
        for i, c := range []struct {
                P,E string
                Matches bool
        }{
                {
                        // Basic match
                        `{"likes":["tacos"],"needs":[{"exists":false}]}`,
                        `{"likes":"tacos"}`,
                        true,
                },
                {
                        // Basic non-match
                        `{"likes":["tacos"],"needs":[{"exists":false}]}`,
                        `{"likes":"tacos","needs":"queso"}`,
                        false,
                },
                {
                        // False, empty array
                        `{"likes":["tacos"],"needs":[{"exists":false}]}`,
                        `{"likes":"tacos","needs":[]}`,
                        true,
                },
                {
                        // True, empty array
                        `{"likes":["tacos"],"needs":[{"exists":true}]}`,
                        `{"likes":"tacos","needs":[]}`,
                        false,
                },
                {
                        // False, explicit null
                        `{"likes":["tacos"],"needs":[{"exists":false}]}`,
                        `{"likes":"tacos","needs":null}`,
                        false,
                },
                {
                        // True, explicit null
                        `{"likes":["tacos"],"needs":[{"exists":true}]}`,
                        `{"likes":"tacos","needs":null}`,
                        true,
                }} {

                t.Run("", func(t *testing.T) {
                        q, err := New()
                        if err != nil {
                                t.Fatal(err)
                        }

                        if err := q.AddPattern(1, c.P); err != nil {
                                t.Fatal(err)
                        }
                        xs, err := q.MatchesForEvent([]byte(c.E))
                        if err != nil {
                                t.Fatal(err)
                        }
                        want := 0
                        if c.Matches {
                                want = 1
                        }
                        if n := len(xs); n != want {
                                t.Fatalf("%d: %v %s %s", i, n > 0, c.P, c.E)
                        }
                })
        }

}
```

core_matcher.go
embano1: Got it. To me a RWMutex would have made it more clear then perhaps, depending on the runtime behavior though.

How long and often do you expect the state machine to be locked compared to reads? E.g., in `etcd` this type of MVCC is implemented using RWMutex.

Of course, if a write mutex is required for a long period, this is going to block readers on the whole object.

On the other hand, with just the atomic value in the struct I'm struggling to understand the benefit of the additional lock then.

Perhaps this one is worth a benchmark comparison? Even if both implementations are close, readability and following the code is easier using one type of atomic operation (e.g. RWMutex) vs atomic and Mutex types combined. At least for me it's harder to understand what's going on with these two types of atomic operations in the same struct.

embano1:
I would love to see some concurrency tests here as well (going back to earlier comment).

E.g. spin up couple hundred of goroutines to simulate reads and writes on the same API and verify correctness for these atomic operations.

Some inspiration for concurrency testing as a start: https://github.com/embano1/memlog/blob/main/memlog_test.go#L433-L469

jsmorph:
> I would love to see some concurrency tests for these edge cases as well (going back to earlier comment).

I've had some clumsy tools lying around. I should tidy things up. Maybe see this sketch of [`quaperf`](https://github.com/jsmorph/quamina/tree/perf/cmd/quaperf), which does an unchecked concurrency test based on generated events and patterns. Might eventually be useful for something.

Many, many caveats apply. Just a sketch!

Screenshot 2022-11-06 173723

Screenshot 2022-11-06 172916

timbray:
> Is it worth noting in `PATTERNS.md` how empty arrays work with `"exists"`?

Definitely. However, I think @yosiat wants to change the behavior because he wants `"exists":true` to work with non-leaf nodes, which seems like a reasonable idea to me.

timbray:
> I would love to see some concurrency tests for these edge cases as well (going back to earlier comment).

The existing concurrency tests (mostly by @jsmorph) were very helpful, actually found a bunch of problems with concurrent access to `map[]` structures (which produces and instant panic) and to slices, which eventually produces an invalid-index fault. I actually think there is now scope for simplification in the `addPattern` code path, it's updating all these arrays by copying them and I think that may no longer be necessary because of the `updateable` fields updated via `atomic.Value`.

But I agree that the edge cases would be a good place to further exercise with concurrency.

embano1:
Regarding the concurrency tests I mentioned, perf tests are always +1 !

Though here I actually meant creating more tests to detect data races based on the introduced changes.

kylemcc:
Just want to say thanks for the hard work on this!

core_matcher.go
timbray: I have benchmarked this one quite a bit.  I'm super-impressed at the performance of atomic.Value - if there is no updating going on, the performance cost is very close to zero. And if I run an updating thread, the slow-down is linear and moderate. See concurrency_test.go.   I had another test there that steadily increased the frequency of updates but it didn't show anything surprising. 

Fortunately each individual addPattern is pretty short  so if you have a bunch of threads trying to put in their own updates it still works pretty well.  I did some reading and found that the discussion of RWMutex in the community discussions was full of caveats and gotchas, so for the moment I like the current lock & atomic.Value design.

timbray:
> spin up couple hundred of goroutines to simulate reads and writes on the same API and verify correctness for these atomic operation

We do that to a certain extent in concurrency_test.go and race_test.go.  They found a lot of bugs and led to the atomic.Value implementation.  It took a lot of iterated executions to expose the  bugs.

timbray:
> Is it worth noting in `PATTERNS.md` how empty arrays work with `"exists"`?

Yes. Doing that now. Then I'd like to push this unless people seriously think we need considerably more concurrency testing.

timbray:
Once again, I'd like to merge this now. Review has improved it, it fixes a nasty bug (adding useful unit tests), and AFAICT the concurrency story is at least no worse than it was before.  An approval would be welcome.

PATTERNS.md
```diff
@@ -66,6 +66,40 @@ are two Exists Patterns that would match the Events above:
...
+```json
+{ "a": [ {"exists": true} ] }
+```
+
+We may be able to change this in future.
+
+The case of empty arrays is interesting. Consider this event:
+
+```json
+{ "a": [] }`
```
embano1, line 87: Is the backtick after `}` supposed to be in here?

embano1:
One nit on the docs.

PATTERNS.md
timbray: Oops good catch.

README.md
```diff
@@ -138,7 +138,7 @@ have corner cases; details are covered in
 Number matching is weak - the number has to appear 
 exactly the same in the Pattern and the Event. I.e.,
 Quamina doesn't know that 35, 35.000, and 3.5e1 are the
-same number. There's a fix for this in the code which 
+same number. There's a fix   for this in the code which
```
embano1, line 141: nit: whitespace but still approving.

embano1 (approved):

timbray:
Thanks. Heh, I forgot -s on one of the commits so I had to force anyhow.

PR #177, opened 2023-01-19
kaizen: Update config for kaizen

timbray:
This was supposed to have been part of the big v1.0-prepare PR but I fat-fingered it out of the commit when fighting git problems co-operatively with Micha.


embano1 (approved):
Added a another `fixup` commit. Works now.

PR #178, opened 2023-01-20
Update CI Workflows

embano1:
- [x] Simplify unit tests using matrix
- [x] Drop go:build tags on ruler tests as they seem unneeded now

Signed-off-by: Michael Gasch <15986659+embano1@users.noreply.github.com>

### DIFF
--- a/.github/workflows/go-unit-tests.yaml
+++ b/.github/workflows/go-unit-tests.yaml
@@ -16,8 +16,15 @@ jobs:
     name: Unit Tests
     strategy:
       matrix:
-        go-version: ["1.18"]
+        go-version: ["1.19"]
         platform: ["ubuntu-latest"]
+        type: ["Tests","Cover"] # run coverage as separate job w/out -race to avoid killing process
+        include:
+          - type: "Tests"
+            goflags: '-v -race -count=1 -json'
+          - type: "Cover"
+            goflags: "-v -count=1 -json"
+            coveropts: "-coverprofile=coverage.txt -covermode=atomic"
 
     runs-on: ${{ matrix.platform }}
     timeout-minutes: 20
@@ -46,10 +53,21 @@ jobs:
       - name: Install tparse
         run: go install github.com/mfridman/tparse@latest
 
+      - name: Check for .codecov.yaml
+        id: codecov-enabled
+        uses: andstor/file-existence-action@20b4d2e596410855db8f9ca21e96fbe18e12930b
+        with:
+          files: .codecov.yaml
+
       - name: Test
         env:
-          GOFLAGS: "-v -race -count=1 -json"
-        run: go test ./... | tparse -all
+          COVER_OPTS: ${{ matrix.coveropts }}
+          GOFLAGS: ${{ matrix.goflags }}
+        run: go test $COVER_OPTS ./... | tparse -all -notests -format markdown >> $GITHUB_STEP_SUMMARY
+
+      - if: steps.codecov-enabled.outputs.files_exists == 'true'
+        name: Upload Codecov Report
+        uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
 
       - name: Verify git clean
         shell: bash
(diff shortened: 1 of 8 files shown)


embano1:
TODO:

- [ ] Update branch protection rules to use the updated unit test name

timbray (dismissed):

.github/workflows/go-unit-tests.yaml
```diff
@@ -18,6 +18,15 @@ jobs:
       matrix:
         go-version: ["1.18"]
         platform: ["ubuntu-latest"]
+        tests: ["NoRuler","Ruler","Cover"] # run those as separate job to avoid killing worker #TODO <issue>
+        include:
+          - tests: "NoRuler"
```
timbray, line 23: I renamed the Ruler test funds, checked these flags, and they seem to work as expected. Now we find out if GitHub can run them.  I considered pulling this fancy matrix into my rename-funcs PR but decided against it because if there's some stupid little syntax problem or other bug you'd be better equipped to fix it.

embano1:
@timbray rebased onto #179 and did three runs, all successful: https://github.com/timbray/quamina/actions/runs/3973740675/usage

embano1:
Update: I made some more modifications, trying to run the `TestRuler` tests together with the regular unit tests. To my surprise, it works. See both runs: https://github.com/timbray/quamina/actions/runs/3974871446

We still have to run `cover` separately due to issues with the `race` detector it seems (tests still time out/get terminated).
But even in the `cover` target, I run `TestRuler` tests (-tags=fromRuler) and they work.

Perhaps we don't need the build tags anymore then?

(disclosure: I updated CI to use Go 1.19, which might have a positive effect here too)

timbray:
Um, if you're running them all together and now it works, that *has* to be go 1.19, right? Unless something else changed?

But if you're convinced that the build tags are no longer necessary, feel free to pull them out as part of this  PR.

Hmm, it's now 9:19AM Pacific time and it looks like the unit tests are somehow blocked from finishing? Looks like 3 hours since last commit?

timbray:
It also occurs to me that I would like to steal a lot more tests from Ruler in the future. Maybe that is an argument for giving them their own spot in the matrix?

embano1:
> Um, if you're running them all together and now it works, that _has_ to be go 1.19, right? Unless something else changed?

It worked with v1.18 too (just used latest in last commit). Not sure what else changed, happy it works :)

> But if you're convinced that the build tags are no longer necessary, feel free to pull them out as part of this PR.

Yup.

> Hmm, it's now 9:19AM Pacific time and it looks like the unit tests are somehow blocked from finishing? Looks like 3 hours since last commit?

Please change this in branch protection rules (new workflow run name). Might have to wait until this is merged.

embano1:
> It also occurs to me that I would like to steal a lot more tests from Ruler in the future. Maybe that is an argument for giving them their own spot in the matrix?

Not sure that's an argument tbh.

timbray (commented):
If you're going to remove the build tags, maybe flip the func names back in those two files? Along the lines of

s/^func TestRuler/func Test/

Or I could just roll back #178 since it seems to be useless now.

embano1:
> If you're going to remove the build tags, maybe flip the func names back in those two files? Along the lines of
> 
> s/^func TestRuler/func Test/
> 
> Or I could just roll back #179 since it seems to be useless now.

Actually I like the naming as it indicates `Ruler`-related tests, no?

timbray:
> Actually I like the naming as it indicates `Ruler`-related tests, no?

OK, not fussy.  So, LGTM. I wonder why the problem went away?

timbray (approved):
Assuming you know how to fix up the branch protection so the unit test completes properly.

timbray:
Oh, still seeing a not-signed problem :(

embano1:
> Oh, still seeing a not-signed problem :(

Haven't added ssh signing. Shall I or merge?

timbray:
> > Oh, still seeing a not-signed problem :(
> 
> Haven't added ssh signing. Shall I or merge?

Go ahead. But let's please set up signing post-1.0?

timbray:
> > > Oh, still seeing a not-signed problem :(
> > 
> > 
> > Haven't added ssh signing. Shall I or merge?
> 
> Go ahead. But let's please set up signing post-1.0?

Go ahead and merge, I meant.

PR #182, opened 2023-01-24
kaizen: vanity import path

timbray:
Worth mentioning that the path is: quamina.net/go/quamina

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -29,7 +29,7 @@ jobs:
         id: go
 
       - name: Run benchmark
-        run: go test  -benchmem -run="^$" -bench "^Benchmark" . github.com/timbray/quamina | tee output.txt
+        run: go test  -benchmem -run="^$" -bench "^Benchmark" . quamina.net/go/quamina | tee output.txt
 
       - name: Download previous benchmark data
         uses: actions/cache@58c146cc91c5b9e778e71775dfe9bf1442ad9a12
--- a/example_test.go
+++ b/example_test.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/timbray/quamina"
+	"quamina.net/go/quamina"
 )
 
 const userRegisteredEvent = `{
--- a/external_test.go
+++ b/external_test.go
@@ -3,7 +3,7 @@ package quamina_test
 import (
 	"testing"
 
-	"github.com/timbray/quamina"
+	"quamina.net/go/quamina"
 )
 
 type fakeFlattener struct {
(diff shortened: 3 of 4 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         id: go
 
       - name: Run benchmark
-        run: go test  -benchmem -run="^$" -bench "^Benchmark" . github.com/timbray/quamina | tee output.txt
+        run: go test  -benchmem -run="^$" -bench "^Benchmark" . quamina.net/go/quamina | tee output.txt
```
embano1, line 32: Question: do we need the go mod path here at all? I think this is enough:

```suggestion
        run: go test  -benchmem -run="^$" -bench "^Benchmark" . | tee output.txt
```

Or do you remember why the module path was added there?

embano1 (approved):
Just one question on the benchmark, otherwise LGTM

embano1:
@timbray to fix the "out-of-date" warning please use this (non-default) option

image

Quiet nice and easy instead of doing this locally.

embano1:
Question: Your current redirect returns:

```console
<meta name="go-import" content="quamina.net/go/quamina git https://github.com/timbray/quamina">
```

IMHO we should follow this guidance:

```
When using modules, an additional variant of the go-import meta tag is recognized and is preferred over those listing version control systems. That variant uses "mod" as the vcs in the content value, as in:
https://pkg.go.dev/cmd/go#hdr-Remote_import_paths
```

.github/workflows/benchmarks.yml
timbray: No idea, sorry. This was done by @yosi I think?

timbray: I want to get this merged so I think I'll leave this in for now post a question to @yosiat and see if we can take this out at some future point

timbray: (I mean, the benchmarks tests are passing and previously they weren't so I want to be careful)

timbray:
> IMHO we should follow this guidance:
> 
> ```
> When using modules, an additional variant of the go-import meta tag is recognized and is preferred over those listing version control systems. 
> That variant uses "mod" as the vcs in the content value, as in:
> 
> <meta name="go-import" content="example.org mod https://code.org/moduleproxy">
> This tag means to fetch modules with paths beginning with example.org from the module proxy available at the URL https://code.org/moduleproxy. See https://golang.org/ref/mod#goproxy-protocol for details about the proxy protocol.

But that would imply that quamina.net should implement the GOPROXY protocol, and have to respond to five different endpoints. So I would actually have to build a real website there at quamina.net so that it would track versions and provide lists and so on.  If the current simplified form actually works it's 

Or am I missing something?


> 
> Source: https://pkg.go.dev/cmd/go#hdr-Remote_import_paths
> ```

Issue #62, opened 2022-06-07
pat: Implement prefix pattern


timbray:
Note that the following are equivalent
```json
{ "a": [ {"shellstyle": "foo*"} ] }
```
```json
{"a": [ {"prefix": "foo"} ] }
```

It's already special-cased in makeShellStyleAutomaton.  So there are a variety of ways this could be done, ranging from just re-writing the pattern to fixing up the switch on ValueType in AddPattern. But no new automaton construction should be required.

PR #190, opened 2023-01-28
pat: add prefix pattern

timbray:
fixes: #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,14 @@ The following Patterns would match it:
   }
 }
 ```
+```json
+{
+  "Image": {
+    "Thumbnail": {
+      "Url": [ "a", { "prefix": "https:" } ] }
+  }
+} 
+```
 The syntax and semantics of Patterns are fully specified
 in [Patterns in Quamina](PATTERNS.md).
 
--- a/pattern.go
+++ b/pattern.go
@@ -217,7 +217,11 @@ func readPrefixSpecial(pb *patternBuild, valsIn []typedVal) (pathVals []typedVal
 		err = errors.New("value for 'prefix' must be a string")
 		return
 	}
-	pathVals = append(pathVals, typedVal{vType: prefixType, val: `"` + prefixString + `"`})
+	val := typedVal{
+		vType: prefixType,
+		val:   `"` + prefixString + `"`,
+	}
+	pathVals = append(pathVals, val)
 
 	// has to be } or tokenizer will throw error
 	_, err = pb.jd.Token()
--- a/pattern.go
+++ b/pattern.go
@@ -19,6 +19,7 @@ const (
 	existsFalseType
 	shellStyleType
 	anythingButType
+	prefixType
 )
 
 // typedVal represents the value of a field in a pattern, giving the value and the type of pattern.
@@ -196,9 +197,30 @@ func readSpecialPattern(pb *patternBuild, valsIn []typedVal) (pathVals []typedVa
 		pathVals, err = readExistsSpecial(pb, pathVals)
 	case "shellstyle":
 		pathVals, err = readShellStyleSpecial(pb, pathVals)
+	case "prefix":
+		pathVals, err = readPrefixSpecial(pb, pathVals)
 	default:
 		err = errors.New("unrecognized in special pattern: " + tt)
 	}
+	return
+}
+
+func readPrefixSpecial(pb *patternBuild, valsIn []typedVal) (pathVals []typedVal, err error) {
+	t, err := pb.jd.Token()
+	if err != nil {
+		return
+	}
+	pathVals = valsIn
+
+	prefixString, ok := t.(string)
+	if !ok {
+		err = errors.New("value for 'prefix' must be a string")
+		return
+	}
+	pathVals = append(pathVals, typedVal{vType: prefixType, val: `"` + prefixString + `"`})
+
+	// has to be } or tokenizer will throw error
+	_, err = pb.jd.Token()
 	return
 }
 
--- a/value_matcher.go
+++ b/value_matcher.go
@@ -119,8 +119,6 @@ func (m *valueMatcher) addTransition(val typedVal) *fieldMatcher {
 			newDfa = nfa2Dfa(newNfa)
 		case prefixType:
 			newDfa, nextField = makePrefixAutomaton(valBytes, nil)
-		default:
-			panic("unknown value type")
 		}
 		fields.startDfa = mergeDfas(fields.startDfa, newDfa)
 		m.update(fields)
@@ -153,9 +151,6 @@ func (m *valueMatcher) addTransition(val typedVal) *fieldMatcher {
 			fields.startDfa = newAutomaton
 			m.update(fields)
 			return nextField
-
-		default:
-			panic("unknown value type")
 		}
 	}
 
@@ -182,8 +177,6 @@ func (m *valueMatcher) addTransition(val typedVal) *fieldMatcher {
 		newDfa = nfa2Dfa(newNfa)
 	case prefixType:
 		newDfa, nextField = makePrefixAutomaton(valBytes, nil)
-	default:
-		panic("unknown val type")
 	}
 
 	// now table is ready for use, nuke singleton to signal threads to use it
--- a/value_matcher.go
+++ b/value_matcher.go
@@ -71,6 +71,7 @@ func (m *valueMatcher) transitionOn(val []byte) []*fieldMatcher {
 
 	default:
 		// no dfa, no singleton, nothing to do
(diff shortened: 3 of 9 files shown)


timbray:
Hmm, I guess I should bump version to 1.1.0, this is a new feature.

timbray:
> is identical to
> 
> ```
> {"a": [ { "shellstyle": "foo*"} ] } 
> ```

Actually, I'm an idiot squared. That breaks if there's a `*` in the prefix. OK, am happy with this now, un-draft-i-fying it, would appreciate a review

README.md
```diff
@@ -113,6 +113,14 @@ The following Patterns would match it:
   }
 }
 ```
+```json
+{
+  "Image": {
+    "Thumbnail": {
+      "Url": [ "a", { "prefix": "https:" } ] }
```
embano1, line 120: nit: 
```suggestion
      "Url": [ "a", { "prefix": "http:" } ] }
```

I guess you did that one on purpose to see if I'm reviewing :)

pattern.go
```diff
@@ -196,12 +197,33 @@ func readSpecialPattern(pb *patternBuild, valsIn []typedVal) (pathVals []typedVa
...
 	}
 	return
 }
 
+func readPrefixSpecial(pb *patternBuild, valsIn []typedVal) (pathVals []typedVal, err error) {
+	t, err := pb.jd.Token()
+	if err != nil {
+		return
+	}
+	pathVals = valsIn
```
embano1, line 213: What's the purpose of this assignment? This is not a (deep) copy.

README.md
timbray: Good catch!

pattern.go
```diff
@@ -196,12 +197,33 @@ func readSpecialPattern(pb *patternBuild, valsIn []typedVal) (pathVals []typedVa
...
+
+	prefixString, ok := t.(string)
+	if !ok {
+		err = errors.New("value for 'prefix' must be a string")
+		return
+	}
+	pathVals = append(pathVals, typedVal{vType: prefixType, val: `"` + prefixString + `"`})
```
embano1, line 220: nit: improve readability by assigning `typedVal{vType: prefixType, val: `"` + prefixString + `"`}` a variable and using the var in `append`

pattern.go
```diff
@@ -196,12 +197,33 @@ func readSpecialPattern(pb *patternBuild, valsIn []typedVal) (pathVals []typedVa
...
+
+	// has to be } or tokenizer will throw error
+	_, err = pb.jd.Token()
+	return
```
embano1, line 224: In general I don't like named returns in Go functions (readability). Any reason we need named returns here?

value_matcher.go
```diff
@@ -116,8 +117,8 @@ func (m *valueMatcher) addTransition(val typedVal) *fieldMatcher {
 			var newNfa *smallTable[*nfaStepList]
 			newNfa, nextField = makeShellStyleAutomaton(valBytes, nil)
 			newDfa = nfa2Dfa(newNfa)
-		default:
```
embano1, line 119: Why did you remove all the `default`s?

value_matcher.go
```diff
@@ -183,6 +187,28 @@ func (m *valueMatcher) addTransition(val typedVal) *fieldMatcher {
...
 }
 
+func makePrefixAutomaton(val []byte, useThisTransition *fieldMatcher) (*smallTable[*dfaStep], *fieldMatcher) {
+	var nextField *fieldMatcher
+	if useThisTransition != nil {
+		nextField = useThisTransition
+	} else {
+		nextField = newFieldMatcher()
+	}
+	return onePrefixStep(val, 0, nextField), nextField
```
embano1, line 197: Nit: can be simplified (avoid `if/else` if possible)

```suggestion
	nextField := newFieldMatcher()
	if useThisTransition != nil {
		nextField = useThisTransition
	}
	return onePrefixStep(val, 0, nextField), nextField
```

value_matcher.go
```diff
@@ -183,6 +187,28 @@ func (m *valueMatcher) addTransition(val typedVal) *fieldMatcher {
...
+}
+
+func onePrefixStep(val []byte, index int, nextField *fieldMatcher) *smallTable[*dfaStep] {
+	var nextStep *dfaStep
+
+	// have to stop one short to skip the closing "
+	if index == len(val)-2 {
+		nextStep = &dfaStep{table: newSmallTable[*dfaStep](), fieldTransitions: []*fieldMatcher{nextField}}
+	} else {
```
embano1, line 206: Nit: simplify, same as above

pattern.go
timbray: Naked returns at work. More on this below

pattern.go
timbray: I confess that this is the one area where I disagree with the Go community; as a result, there are named returns (heh, "naked" returns) here and there around Quamina. The linter doesn't seem to care.

For the moment, I don't feel any urgency about chasing them out. If someone cared enough I'd approve a PR that takes them all out systematically.

value_matcher.go
timbray: But wouldn't that cause a wasted call to `newFieldMatcher` in the (common) `!= nil` case?

value_matcher.go
timbray: Same response; wouldn't that cause a wasted memory allocation?

pattern.go
embano1: Ok, besides the named (naked) returns, you don't care that this is a shall copy of `valsIn` here, i.e. immutability for the source slice is not required, correct?

pattern.go
embano1: This is what I meant and I feel it improves readability, but this is nit:

```go
val := typedVal{
	vType: prefixType,
	val:   fmt.Sprintf("%q", prefixString), // or keep the concat if you're concerned about allocs
}
pathVals = append(pathVals, val)
```

value_matcher.go
```diff
@@ -71,6 +71,7 @@ func (m *valueMatcher) transitionOn(val []byte) []*fieldMatcher {
 
 	default:
 		// no dfa, no singleton, nothing to do
+		// this probably can't happen because a flattener shouldn't preserve a field that hasn't appeared in a pattern
```
embano1, line 74: nit: comment line wrapping seems off

value_matcher.go
embano1: You could test it IMHO (if you're concerned about coverage): create a new `valType` in your test and pass it. It should trigger the panic (which you can recover from and assert). Reason I like `panic` in those `switch` statements is that with the code base growing, slipping in a new `valType` and forgetting the places where this is `switch`-ed on is a possibility. Without a `default` case your code would happily continue and likely produce a bug.

value_matcher.go
embano1: Was wondering if the compiler would be smart enough to figure out that branch and avoid the heap alloc as well. Looks like we don't have benchmarks here so can't judge. Is this a hot path or more like a one-off at startup?

value_matcher.go
embano1: Update: here's an example how to start testing these cases

```go
func TestAddInvalidTransition(t *testing.T) {
	defer func() {
		if recover() == nil {
			t.Errorf("TestAddInvalidTransition should have panicked")
		}
	}()

	panicType := valType(999)

	m := newValueMatcher()
	v1 := typedVal{
		vType: panicType,
		val:   "one",
	}
	m.addTransition(v1)
}
```

Note: you could add more panic scenarios to cover the `if fields.startDfa != nil` branch. `default` in `singletonAutomaton` might not be needed as caught by earlier panic.

embano1:
Just nits left, I'll approve and let you decide whether you want to make small adjustments based on the review. At least reverting the `default` scenario seems like a zero-waste but good engineering practice.

embano1 (approved):

value_matcher.go
timbray: Excellent, thanks, will do.

value_matcher.go
timbray: Actually, it turn out that the useThisTransition has become a one-off special case and in fact the argument is going to be removed from most of the build*Automaton calls. So I'm going to leave this as is for now and address in a future kaizen PR.

value_matcher.go
timbray: as above

pattern.go
timbray: OK, you're right, it adds readability.

value_matcher.go
timbray: This definitely improved both the code and the coverage, thanks.

embano1: Awesome, glad to hear!

embano1 (approved):
LGTM

PR #259, opened 2023-12-08
Issue 252

timbray:
The problem was that `{"data": {"field": []}}`, which is syntactically valid, would make `addPattern` blow up.  Which at one level is sane because it can't match anything, but on the other hand if you're generating Patterns it would be easy to write code to "match any of the elements of my array `PartNumbers`" and if `PartNumbers` was empty you'd get this kind of pattern and while it couldn't match anything, that would be the expected correct behavior.

The reason for excluding that particular message in the gosec linter is that the linter wants the crypto/rand as opposed to math/rand, but we want to retain the use of a fixed seed for the random number generator for testing/debugging purposes, and crypto/rand doesn't seem to support that and also is slower.

### DIFF
--- a/pattern_test.go
+++ b/pattern_test.go
@@ -41,6 +41,60 @@ func TestPatternErrorHandling(t *testing.T) {
 	}
 }
 
+// test that adding an empty pattern doesn't screw anything up
+func TestEmptyValueArray(t *testing.T) {
+	var err error
+	empties := []string{
+		`{"data": {"field": []}}`,
+		`{"data": {"field": [], "field2": [23]}}`,
+		`{"data": {"field2": [], "field": [23]}}`,
+	}
+	wanted := map[string][]X{
+		`{"data": {"field": 23}}`:  {empties[2]},
+		`{"data": {"field2": 23}}`: {empties[1]},
+		`{"data": {"field": []}}`:  {},
+	}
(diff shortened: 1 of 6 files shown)


pattern_test.go
```diff
@@ -41,6 +41,61 @@ func TestPatternErrorHandling(t *testing.T) {
...
+
+	for want := range wanted {
+		matches, err := cm.matchesForJSONEvent([]byte(want))
+		if err != nil {
+			t.Errorf("m4je err on %s", want)
+		}
+		checkXEqual(t, wanted[want], matches)
+	}
+}
+func checkXEqual(t *testing.T, x1s []X, x2s []X) {
```
embano1, line 78: nit: I just wonder whether `slices.Sort` and then `slices.Equal` from https://pkg.go.dev/golang.org/x/exp/slices would be more clear instead of creating a new equality check helper?

embano1 (approved):
One nit/question.

Also, can you please update your PR description. Currently it only contains the linter issue but this PR contains also the fix for empty array.

pattern_test.go
timbray: Good idea. Unfortunately the "X" type doesn't implement cmp.ordered.

timbray: Updated PR

Issue #197, opened 2023-02-18
kaizen: Cleanup DFA/NFA


timbray:
I am starting to implement the "suffix" pattern which obviously requires an NFA to implement (as does "shellstyle" and anything slightly regexp-ish). At the moment, Quamina reduces all NFAs to DFAs, which is well-known to be prone to producing O(2**N) explosions of the kind we're seeing in some tests (although it's not obvious this is the source of that problem).  Anyhow, the DFA/NFA confusion is adding lots of complexity. We should throw away DFAs and do everything in NFAs.  This would have a few benefits, including:

- lose a lot of code
- simplify the addPattern code path
- de-genericize the SmallTable structure - this doesn't seem to be a good use of generics, lots of things are made harder to read and write.

timbray:
Fixed in [8833629](https://github.com/timbray/quamina/commit/883362999b5b31baa194be3e21bfac56cfa3fc99)

PR #308, opened 2024-05-25
kaizen: clean up state finite automata

### LINKED ISSUE #188: fix: state-machine size explosion on wildcards
Certain combinations of `shellstyle` patterns cause an explosion in the size of the state machine along the lines of O(2**N).  The effect is that `AddPattern()` can take seconds to execute and burn memory.  We know this is not inevitable because AWS Event Ruler can build this kind of machine quickly and economically.

Initial thoughts:
1. Characterize the problem properly
2. Make nice `String()` on smallTable and related types
3. Replace `list_maker` with something much better thought through - it's a hack
4. Fix the problem

timbray:
addresses #197

This gets rid of the DFA/NFA distinction, everything is now an NFA. The stretch goal is to fix #188, but the messy sprawl of DFA-and-NFA was getting in the way of understanding. 

This includes a bunch of commented-out code for pretty-printing state machines, which I found totally essential in migrating to all-NFA.  I'm going to create an issue to bring this code out of comments and make it possible to switch it off/on to facilitate the creating and merging different kinds of patterns.


timbray:
I see that I accidentally lost a couple of unit tests, although the coverage didn't go down.  Will clean this up subsequently.  There's a huge amount of quite difficult work here that I want to get checked in so I have a stable base to continue work on #188 and then some new features.

timbray:
Having thought about it, I can see why this is a difficult PR to review.  It's all down in the finite-automaton weeds and not that many people care about this stuff or have exposure to it.  Having said that, a bunch of ugly code was removed and the tests all still pass and I see no performance regression.  So I will probably break the rules and go ahead and merge if nobody is up to doing a real review in the next day or so.

embano1:
+1

Issue #310, opened 2024-05-31
kaizen: Add a pretty-printer


timbray:
Building and debugging NFAs is hard. I built a pretty-printer for the DFA/NFA cleanup PR but commented it out, it should be rolled into the source code, have a test suite, etc.

timbray:
Done in #311

PR #311, opened 2024-05-31
addresses issue #310

timbray:
add a prettyprinter for NFAs


timbray:
Once again, deep NFA twiddling, will push it myself if nobody wants to review.  However, have a glance at TestPP() in prettyerinter_test.go to see what the prettyprinter output looks like for the wildcard pattern "x*9".  I found this output absolutely essential in debugging the transition from DFA to NFA. In that output, the abbreviation `gS` stands for "glob state", i.e. when you arrive at `*` and loop back on anything but a particular byte value. The state you reach when you see that particular value and exit the glob has a name including `gX` for "glob exit".

If you have an idea for an improvement in the prettyerinter output, feel free to comment.

prettyprinter.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"fmt"
+	"math/rand"
+	"strings"
+)
+
+// printer is an interface used to generate representations of Quamina data structures to facilitate
+// debugging and optimization. It's an interface rather than a type so that a null implementation can
+// be provided for production that should incur very little performance cost.
+type printer interface {
```
embano1, line 12: I'd suggest following what Google did when they developed the `slog` logging package. Instead of going with an `interface` type, they used a concrete struct with a specific `interface` handler type to avoid allocations:  https://youtu.be/8rnI2xLrdeM?feature=shared&t=1336 (starts at that conversation)

field_matcher.go
```diff
@@ -93,7 +93,7 @@ func (m *fieldMatcher) addExists(exists bool, field *patternField) []*fieldMatch
 	return []*fieldMatcher{trans}
 }
 
-func (m *fieldMatcher) addTransition(field *patternField) []*fieldMatcher {
+func (m *fieldMatcher) addTransition(field *patternField, printer printer) []*fieldMatcher {
```
embano1, line 96: See my comment on using a concrete `nullOp` by default struct for printing to avoid allocations.

prettyprinter.go
embano1: tl;dr

```go
// inspired by https://pkg.go.dev/golang.org/x/exp/slog#Logger
type Handler interface {
  Enabled() bool
  Handle(...)
}

type Printer struct {
  handler handler
}

func New() *Printer {...}
```

embano1 (commented):
Left a comment regarding potential performance (allocation) issues. You should also update the README if you think this could be useful. For example, would it be good to enable printing behavior (for users) using an environment variable like `DEBUG_NFA` to enable the printer?

Regarding printing output, might also be worth explaining that in README unless you want users/developers to only send a printer output file to you so you can debug?

prettyprinter.go
timbray: Also, I agree about documenting this stuff. Probably best to create a `DEVELOPING.md` or some such; is there a standard name for this kind of thing?

embano1: If this implementation is purely for developing, `CONTRIBUTING.md` might be common

core_matcher.go
```diff
@@ -52,6 +52,12 @@ func (m *coreMatcher) fields() *coreFields {
 // addPattern - the patternBytes is a JSON text which must be an object. The X is what the matcher returns to indicate
 // that the provided pattern has been matched. In many applications it might be a string which is the pattern's name.
 func (m *coreMatcher) addPattern(x X, patternJSON string) error {
+	return m.addPatternWithPrinter(x, patternJSON, &nullPrinter{})
+}
+
+// addPatternWithPrinter can be called from debugging and under-development code to allow viewing pretty-printed
+// NFAs
+func (m *coreMatcher) addPatternWithPrinter(x X, patternJSON string, printer printer) error {
```
embano1, line 60: since this is taking in an `interface` type (`printer`), IMHO that is the allocation/escaping issue raised in the Youtube link I shared, no?

prettyprinter.go
embano1: Left a comment in the place where I think alloc might happen? Ideally, we'd have a benchmark which shows the before/after here?

embano1: Can't see alloc changes in this though: https://github.com/timbray/quamina/actions/runs/9324906941?pr=311

core_matcher.go
timbray: Ahh, I get it. I confess, I didn't watch the YouTube, I just went for obvious allocation.  Now I have, and I learned all sorts of stuff.  In particular `go build -gcflags=-m` and `treating.AllocsPerRun`. Thanks for that!

Now… it’s complicated.  If we turn printer into a struct type where all the methods start 
```go
if !pp.enabled() {
  return constant;
}
// else do real work
```
then…
1. there is one less allocation for any method that has a `printer` argument
2. each ``printer.anything()`` call actually executes the conditional above
3. but it's a direct call.

On the other hand, if we use the current interface approach:
1. There is an allocation wherever the method has a printer argument.
2. but in the nullPrinter case there's no conditional executed, it just does `return`.
3. but it's a virtual call so runtime has to do a lookup

Other factors…
1. This is only called on `addPattern` calls which are much rarer than `patternsForEvent` calls
2. The `addPattern` call is fairly expensive, does lots of computation and allocation
3. The allocation is about the minimum possible, an empty structure and dispatch information

The two most significant factors in my mind are both intuitions unsupported by data:
1. The effect is small and will not  be significant in the context of the heavyweight `addPattern` logic
2. Of the effects, the struct call vs the indirect interface call might actually be a big deal.

What do you think?

prettyprinter.go
timbray: Yes, CONTRIBUTING is the place.

core_matcher.go
embano1: Thx for clarifying that this is only on the addPattern code path - missed that. Then I completely agree that interface approach works bc we are not in the data path 👍

prettyprinter.go
timbray: I added a description to CONTRIBUTING in the latest commit. If that's OK I think we can resolve this conversation?

CONTRIBUTING.md
```diff
@@ -99,3 +105,52 @@ instructions for installing it.
 
 When opening a new issue, try to roughly follow the commit message format
 conventions above.
+
+## Developing
```
embano1, line 109: Great section! Since we talked about OS env vars, how would a dev enable the pretty printer easily? Would be nice to explain this here too.

timbray: Good idea. CONTRIBUTING now contains a mini developer's user guide.

embano1 (approved):
LGTM, perhaps squash before merging

timbray:
> LGTM, perhaps squash before merging

Thanks! Your input really improved this one.  I'll let GitHub do the squashing, that seems to work OK with the CI.

PR #320, opened 2024-06-18
pat: remove limit of one * in shell style

timbray:
Note that this is a semantic change so will lead to release 1.1.0.


timbray:
The CI is breaking because "package slices is not in GOROOT".  I just introduced slices in a test so I could use slices.Contains.  I could easily take slices out and write my own "contains" method in about 3 minutes but I'm reluctant to do that without understanding what's going on. Is the "slices" package magically weird?

timbray:
OK, fairly obvious that nobody wants to review PRs that dive deep into finite-automaton wrangling. Will merge.  Next PR is about getting numeric matching right, which at least is user-visible.

PR #330, opened 2024-07-08
pat: Correct number matching (again)

timbray:
Thanks to @jsmorph for concurrency_test.go, whose surprising result revealed that the work for this PR was not nearly finished. More work was required on (a) detecting whether a number in an incoming event could be made comparable, (b) communicating that fact to the matching code, (c) marking NFAs which contained comparable numbers, leading to the following statement in value_matcher.go:
```
		if vmFields.hasQNumbers && eventField.IsQNumber {
			qNumber, err := qNumFromBytes(val)
```
Also, (d) the unit tests needed work to cover the cases that were previously issued.

Now, there is still a performance cost, particularly in benchmarks that do a lot of matching of numbers; correctness isn't free. But at the moment, I think the trade-off is acceptable.

Note that having previously called these things `comparableNumbers` and `canonicalNumbers` and disliking both and getting long ugly variable names, I renamed them to Q numbers, where Q is for Quamina.


timbray:
tl:dr: I'm OK with the slowdown reported here.

As remarked, the numeric processing cost is real.  This benchmark that shows this is using the pattern:
```json
{ "context": { "user_id": [9034], "friends_count": [158] } }
```
Which is purely numeric. So, instead of matching 3- and 4-byte numbers, we're matching 14-byte canonical representations.  Interestingly, when I measure matches/second metrics, the slowdown is dramatically smaller than that reported by the GitHub CI benchmark runner. (Not sure why.)

Finally, for the particular type of NFA produced for number atching, there are several obvious optimizations in the way the NFA-matcher runs, which will further reduce the impact of numeric accuracy.  I think accuracy is important, so that an event like the following would match the pattern above.
```json
{ "context": { "user_id": 9034, "friends_count": 1.58e2 }} 
```
(or `158.0` or any other variant)

PR #339, opened 2024-07-20
chore: drop Quamina.flattenerSpecified

arnehormann:
The information provided by flattenerSpecified is redundant and available by checking if flattener is nil.

### DIFF
--- a/quamina.go
+++ b/quamina.go
@@ -11,7 +11,6 @@ import (
 type Quamina struct {
 	flattener          Flattener
 	matcher            matcher
-	flattenerSpecified bool
 	mediaTypeSpecified bool
 	deletionSpecified  bool
 }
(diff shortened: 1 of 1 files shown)


quamina.go
```diff
@@ -101,7 +99,7 @@ func New(opts ...Option) (*Quamina, error) {
 			return nil, err
 		}
 	}
-	if !(q.mediaTypeSpecified || q.flattenerSpecified) {
+	if !q.mediaTypeSpecified && q.flattener == nil {
```
timbray, line 102: Could we please fully parenthesize this kind of condition?  As a matter of principle I never remember the precedence rules in any programming languages, which forces me to put in a few unnecessary parentheses, which I suspect has saved me from disasters more than once. A
```go
if (!q.mediaTypeSpecified) && (q.flattener == nil)

timbray (commented):
LGTM. One small code nit

quamina.go
arnehormann: done

timbray (approved):
Thanks!

Issue #186, opened 2023-01-26
pat: Add equals-ignore-case matching


timbray:
Syntax/semantics as defined for AWS Event Ruler

timbray:
I thought "this will be an easy little win" and started working on it, but it's a forest of ugly details.  I also think that it's also in important respects just wrong (case-folding semantics are very locale-specific) but let's stay compatible with Ruler.

timbray:
Added in #340

PR #340, opened 2024-07-20
kaizen: add equals-ignore-case pattern

timbray:
addresses: #186

This adds support for the `equals-ignore-case pattern`. The automaton-building is a bit complex but reasonably well tested. 

What is controversial is that this PR contains Quamina's first generated code - the case-mappings are pulled from the file CaseFolding.txt in the Unicode Character Database. The generated code is in `case_folding.go`.  It is built by `code_gen/build_casefolding_table.go` which is in package `main` and has `main()`

The problem is that the CaseFolding.txt is regularly changed with Unicode releases. So `build_casefolding_table` checks to see if the `case_folding.go` file is older than 3 months and if so, rebuilds it.  There is a `Makefile` to take care of building and running the code generator.  So if you're doing a PR you should at some point type `make` and if you get a new `case_folding.go` then check it in with your PR. 

There are a couple of problems. First, I've never done generated-Go code before and maybe there's a much better well-known approach.  Second, I haven't figured out how to unit test the code generator. The coverage tools ignore it because it's not in the quamina package. Implicitly, it's tested by the tests in `monocase_test.go` and elsewhere having the desired results but I'm going to open an issue to figure out how to test properly.


timbray:
Hmm, surprised by the benchmark failures, I didn't think I changed anything that should affect an existing benchmark.  Will investigate. Could be trailing fallout from the Q-numbers work.

arnehormann:
@timbray Go itself (the x package is semi-official and comes close to the standard library) appears to do it this way:
https://cs.opensource.google/go/x/text/+/master:gen.go
with generated code like this: https://cs.opensource.google/go/x/text/+/master:cases/tables15.0.0.go

timbray:
This is annoying; the same problem with the same benchmark showed up on the Q numbers PR, and I accepted the numeric-match slowdown in the interests of correctness. So I changed the CI/CD to warn but not fail, and then after everything was clean I put the CI/CD back to fail mode, assuming that would reset the cached previous value.  Apparently this did not happen; I have poked around and failed (so far) to figure out how/where the benchmark tool caches previous results. 

Anyhow, for now, I'm going to reset the CI/CD again not to fail on slowdown.

PR #349, opened 2024-08-18
numbits+base128 8-byte full-precision numbers

timbray:
For background, see #336 

The effect is ability to match full range of float64 and more compact representation, with significant speedup.


timbray:
BTW there's something wrong with the benchmarks - my local benchmarks which measure matches/second rather than ns/op have slightly improved in this merge. Need to investigate.

numbits.go
```diff
@@ -0,0 +1,41 @@
+package quamina
+
+import (
+	"math"
+)
+
+// float64 are stored as (sign | exponent | mantissa)
+// with 1 bit sign, 11 bits exponent, 52 bits mantissa
+const ()
```
timbray, line 9: Oops, editing error here didn't notice the empty const() and my introduction needs to move up to the top.

timbray:
Thanks again @arnehormann, the variable-length numbits work perfectly. The numbits type is still `long` of course, but its `toUTF8()` method does the shortening.  I've verified that for the “common” numbers, it produces much shorter slices. 

We currently don't have a good benchmark to demonstrate a performance improvement, because all of our number benchmarks are exploring corner cases with weird huge/tiny/high-precision numbers, not just ordinary numbers like the kind that would actually be used. But I'm confident that in practice this will give higher performance.

flattener.go
```diff
@@ -58,5 +58,5 @@ type Field struct {
 	Path       []byte
 	Val        []byte
 	ArrayTrail []ArrayPos
-	IsQNumber  bool
+	isNumber   bool
```
ah-quant, line 61: No change needed, just an api-design question:
Why the change to unexported? It doesn't have to be exported, but since the struct and all other fields and the previous version are all exported, it stands out.
If quamina is used as a library, it could be nice to now and it's probably more relevant for clients than e.g. ArrayTrail.

numbers.go
```diff
@@ -1,121 +1,56 @@
...
-// They include most numbers that are used in practice, including prices, occurrence counts, size
-// measurements, and so on.
-// Examples of numbers that do NOT meet these criteria include AWS account numbers, some telephone
-// numbers, and cryptographic keys/signatures. For these, treatment as strings seems to produce
-// satisfactory results for equality testing.
+// in text data or the internal floating-point bits. Therefore, we map floating-point numbers
+// (which is what JSON numbers basically are) to comparable strings which preserve the ordering
+// numbers' ordering. Versions of Quamina up to 1.3 used a home-grown format which used 14 hex digits
+// to represent a subset of numbers. This has now been replaced by Arne Hormann's "numbits"
+// construct, see numbits.go. It uses 10 base128 bytes to represent the entire range of float64 numbers.
```
ah-quant, line 13: change "10" to "up to 10"?

numbers.go
```diff
@@ -1,121 +1,56 @@
...
 }
 
-func toHexStringSkippingFirstByte(value uint64) []byte {
-	var buf [8]byte
-	binary.BigEndian.PutUint64(buf[:], value)
-	var outputChars [14]byte
-	for i, utf8Byte := range buf {
-		if i == 0 {
-			continue
+/*
```
ah-quant, line 44: drop?

numbits.go
```diff
@@ -0,0 +1,45 @@
...
+	// transform without branching (inverse of numbits.Float64):
+	// if high bit is 0, xor with sign bit 1 << 63, else negate (xor with ^0)
+	mask := (u>>63)*^uint64(0) | (1 << 63)
+	return numbits(u ^ mask)
+}
+
+// toUTF8 turns a numbits into 10 bytes of UTF-8 encoded via Base-256
+// code copied with thanks from a sample by Axel Wagner
+func (nb numbits) toUTF8() []byte {
+	var b [10]byte
```
ah-quant, line 33: If performance critical, this could be unrolled as shown in https://github.com/timbray/quamina/issues/334#issuecomment-2295324890 - both versions do the same big endian base128 encoding.
```
be128 := []byte {
  byte(num >> 57) & 0x7f, // 57 bits remaining
  byte(num >> 50) & 0x7f, // 50 bits remaining
  byte(num >> 43) & 0x7f, // 43 bits remaining
  byte(num >> 36) & 0x7f, // 36 bits remaining
  byte(num >> 29) & 0x7f, // 29 bits remaining
  byte(num >> 22) & 0x7f, // 22 bits remaining
  byte(num >> 15) & 0x7f, // 15 bits remaining
  byte(num >> 8) & 0x7f, //  8 bits remaining
  byte(num >> 1) & 0x7f, //  1 bit  remaining
  byte(num & 1) << 6, // last bit (high bit of lower 7 bits)
}
```
and "math/bits" can be used to slice it: `(bits.TrailingZeros64(uint64(nb)) + 6) / 7` can determine the end without looking.

numbers.go
```diff
@@ -1,121 +1,56 @@
...
-// In English: all numbers that are between negative and positive 5 billion inclusive, with up to five
-// digits after the decimal point.
-// These numbers have fifteen decimal digits of precision, which is what double floats can offer.
-// They include most numbers that are used in practice, including prices, occurrence counts, size
-// measurements, and so on.
-// Examples of numbers that do NOT meet these criteria include AWS account numbers, some telephone
-// numbers, and cryptographic keys/signatures. For these, treatment as strings seems to produce
-// satisfactory results for equality testing.
+// in text data or the internal floating-point bits. Therefore, we map floating-point numbers
+// (which is what JSON numbers basically are) to comparable strings which preserve the ordering
```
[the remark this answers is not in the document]
timbray: Yep. 

Also, thanks a lot for the review.  I lost reviewers when all my PRs were digging around inside the NFA implementation, which is really very difficult to review.

case_folding.go
```diff
@@ -3,475 +3,475 @@ package quamina
 // Code generated by code_gen/code_gen - DO NOT EDIT.
 // built from the "C" records in CaseFolding.txt in the Unicode character database
 var caseFoldingPairs = map[rune]rune{
```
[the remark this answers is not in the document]
timbray: This is a symptom of a git mistake.  The lint checker changed in GitHub but I hadn't updated my local copy, so all of a sudden there were these CI failures.  So I did PR #348  that addressed just the lint problems.  But I neglected that I also had those in my working copy where I was working with numbits. So they showed as a diff in this PR, even though they are already in the  origin repo. I tried to fix with `git pull` but no luck.  So you are correct that this is unrelated but it isn't really in the PR.  If you have advice how to manipulate with git to avoid this, that would be good.

code_gen/build_casefolding_table.go
```diff
@@ -43,21 +43,21 @@ func main() {
 	}
 	resp, err := http.Get(CaseFoldingURL)
 	if err != nil {
-		fatalF("Can't fetch CaseFolding.txt: " + err.Error())
+		fatal("Can't fetch CaseFolding.txt: " + err.Error())
```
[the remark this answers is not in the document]
timbray: Exactly. Added to the commit by mistake. Not sure how to remove.

flattener.go
timbray: Oh, just a mistake, should be exported.

numbers.go
timbray: yes

numbers.go
timbray: I want to keep this around because it's very useful for debugging and I will definitely need it when I'm implementing the range construct.  I guess I should un-comment and use in a test so that I don't get lint complaints about unused function.

numbits.go
timbray: I prefer the loop construct, I think it's roughly as readable and takes less screen space.  Or am I missing something?

As for `TrailingZeros64`, I think there's a problem, which is the case when the numbits reduces to just `[0]` (the smallest possible number I guess).  Because removing all the trailing zeroes would reduce this to a n empty slice and the automaton needs at least one byte to match on. That's why the current loop has `k>0` not `>=0`. But, I should comment that because it's non-obvious.

arnehormann: numbit 0 is NaN, cannot happen in JSON ☺️

timbray: Heh, OK. But I better say that in a comment…

case_folding.go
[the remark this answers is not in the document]
arnehormann: You could do a soft reset, add everything you need selectively and commit then. followed by a force push to overwrite the branch. and do all that on a new local branch to make sure you can still revert back to it if you stumble somewhere. 
But there's a simpler, hacky approach:

git fetch origin/main --prune
git checkout -b tmp origin/main
git checkout numbits -- <list of files you want>
git add <list of files you want>
git commit
git push -f origin tmp:numbits

That's probably nicer to navigate than with revert etc - directly pick the files you want in a new branch instead of wrangling commits.
requires a bit of cleanup because your local numbits branch will be different to the server one and you should get rid of the tmp branch. But I don't want to add the commands to do destructive operations on your local repo.

If you add stuff and you have multiple changes in your repo, add files individually. add -p can also be used to only add regions in a file. And then, take care to not add -a to commit - that will drag everything else back in.

numbits.go
arnehormann: also, base 256 would be a plain byte. it's big endian base 128 (only the lower 7 bits are used for all bytes, byte order is big endian).

timbray: Oh you're right.  Off by 1.

case_folding.go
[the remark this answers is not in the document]
timbray: Since I don't understand what's really going on in that sequence, unless you think it's really damaging, I think I'd prefer to live with the noisy PR. I'm a git idiot.

numbits.go
timbray: Anyhow, I tried out `bits.TrailingZeros64` but the code was a bit awkward and the +6 / 7 nonintuitive, but I did go back and optimize a bit; first count back after 0-byte encodings, and leave everything set up for a separate loop that allocates just the right size and copies them in.

timbray:
Wow, I **really** need to update my local copy of the linter.  Surprisingly, I think that lint failure is an actual bug; I will try to prove that with a unit test.

numbits.go
arnehormann: it still says Base-256 in the comment

arnehormann: I also made a mistake, there. the +6/7 is only needed if we go from left to right. 
It should have been `be128[:len(be128)-bits.TrailingZeros64(uint64(nb))]`

arnehormann:
... and just fyi (expectation management): this variable length scheme will only compress positive integers and power-of-two fractions. It won't do much for negative values because they had to be negated (xored with one bits) in the process to make them lexically sortable.
Still worth it, it doesn't cost much time and no additional bytes at all.

timbray:
Haha, you're unduly pessimistic. numbits are integer-friendly.  For the ints 0 to 100,000 inclusive, here are the counts of shortened-numbits lengths: 
```
1: 1
2: 1
3: 115
4: 7590
5: 92294
```

None higher than 5. All <5 until you get over 1K. But you are correct, negative numbers all seem to be length 10

This is actually very good because in real life, many (most?) of the numbers that get matched are integers, not too large.

arnehormann:
as I said: only positive integers and power of two fractions. but positive values can be as little as 2 bytes 😁

timbray:
BTW on that last commit I neglected to add the most recent version of numbits, which has an optimized shortening that causes a tiny performance improvement. A little short of sleep these days…

PR #371, opened 2024-10-10
improvement: alternative numbits conversion

arnehormann:
... and drop a doc-reference to a function that was not ported from the original implementation.

The improvement will use sign extension for a right shift for int64. This will fill all bits with the sign bit. There's no need for another constant.

This was sensibly suggested by Raph Levien in a mastodon thread: https://mastodon.online/@raph/113071041069390831

### DIFF
--- a/numbits.go
+++ b/numbits.go
@@ -21,6 +21,7 @@ type numbits uint64
 // numbitsFromFloat64 converts a float64 value to its numbits representation.
 func numbitsFromFloat64(f float64) numbits {
 	u := math.Float64bits(f)
+	//nolint:gosec // disable G115
 	// transform without branching:
 	// if high bit is 0, xor with sign bit 1 << 63, else negate (xor with ^0)
 	mask := uint64(int64(u)>>63) | (1 << 63)
--- a/numbits.go
+++ b/numbits.go
@@ -23,7 +23,9 @@ func numbitsFromFloat64(f float64) numbits {
 	u := math.Float64bits(f)
 	//nolint:gosec // disable G115
 	// transform without branching:
-	// if high bit is 0, xor with sign bit 1 << 63, else negate (xor with ^0)
+	// if high bit is 0, xor with sign bit 1 << 63, else negate (xor with ^0).
+	// Using a sign extending right shift was proposed by Raph Levien in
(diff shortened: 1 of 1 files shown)


timbray:
Sorry Arne, am still in the chaos of moving house, I'm sure I'll just accept this when I get free in a week or so.

baldawar:
If it helps, the change matches what ruler has at https://github.com/aws/event-ruler/blob/main/src/main/software/amazon/event/ruler/ComparableNumber.java#L65. This code has been running without any issues in shadow mode on production workloads for almost a month now. I don’t understand the float -> int -> uint conversion but I imagine that’s peculiar to go number types.

numbits.go
```diff
@@ -21,9 +21,10 @@ type numbits uint64
 // numbitsFromFloat64 converts a float64 value to its numbits representation.
 func numbitsFromFloat64(f float64) numbits {
 	u := math.Float64bits(f)
-	// transform without branching (inverse of numbits.Float64):
+	//nolint:gosec // disable G115
+	// transform without branching:
 	// if high bit is 0, xor with sign bit 1 << 63, else negate (xor with ^0)
```
timbray, line 26: I'd like to see a credit to Raph for suggesting this technique.

arnehormann:
@baldawar it's something I knew and _should_ have done, too. One of these "obvious in hindsight" things, like numbits itself:
Right shifting is a division by two to the power of the number of bits.
Due to twos-complement, high bits have to be filled with the first bit to make this work with signed numbers.
Java has no unsigned primitive numeric types, so they provide two kinds of right shift: one that fills new high bits with `0` and one that fills the new high bits with copies of the previous highest bit.
If you have a positive unsigned value with the highest bit set (greater or equal than `1<<63`) or for regular binary operations, you need to use the "fill with zeroes" version (in Java: `>>>`). Go has both signed and unsigned types and has only one way to right shift. The casts are required to make this do the right thing in machine code.

In the code here and in Ruler, it's used to avoid another constant. Which is a teeny tiny bit faster but also looks nicer imo. Algorithmically, at least. In this case, Java definitely looks cleaner than Go due to its own special operator `>>` vs `>>>`.

timbray (approved):
LGTM

Issue #66, opened 2022-06-12
Implement I-Regexp


timbray:
The IETF JSONPath WG is building a minimal-ish regular expression specification called I-Regexp designed to be interoperable across a wide variety of regex implementations. See https://www.ietf.org/id/draft-ietf-jsonpath-iregexp-00.html

A full or even subset implementation of I-Regexp would be a valuable addition to Quamina.

timbray:
As of 2024/09, I'm starting to take a run at this. If we get a substantial chunk of I-Regexp working, some of the other pattern upgrades can be re-implemented as regexes. Am going to do this incrementally, pulling in features one at a time, but have to have a full parser because we should reject regexes that use unimplemented features.

timbray:
BTW I-Regexp is now RFC 9485.

PR #400, opened 2025-01-31
pat: Add char classes and alts to regexp

timbray:
partially address #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,13 @@ The following Patterns would match it:
   }
 }
 ```
+```json
+{
+  "Image": {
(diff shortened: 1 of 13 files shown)


timbray:
After I posted this I discovered, in Go's "unicode" pkg, the `RangeTable` type, which is more or less identical to the `RuneRange` type included here. Since all of Go's built-in tables for `\P` and `\p` formations are expressed in terms of RangeTables, we should drop RuneRange and switch over.  But I want to do that after I get this checked in, will make the change when I address \p and \P.

PR #466, opened 2026-01-16
kaizen: Switch sorting to use the slices package.

happygiraffe:
This was introduced in Go 1.21.

Potentially it has better perfomance as the API allows it to make fewer comparisons. Mostly, I think it reads nicer.


happygiraffe:
Apologies — I've not set up signed commits. I need to go figure out how to do that.

timbray:
1. Yes please if you could arrange to be signed, it's just a matter of pushing a public key IIRC. I'd have to change the CI to bypass that and would prefer not to.
2. Yes, there is a performance improvement.  I have a little basic benchmark that I guess I should check in because I use it all the time, I've attached it to this comment.  Anyhow here is its report.

Before #466 
EXACT: 2.732
ANYTHING-BUT: 0.369
SHELLSTYLE: 1.411
EQUALS_IGNORE-CASE: 1.567
REGEXP: 1.590
Avg: 2.556

After:
EXACT: 2.841
ANYTHING-BUT: 0.369
SHELLSTYLE: 1.480
EQUALS_IGNORE-CASE: 1.652
REGEXP: 1.691
Avg: 2.678

The numbers are in millions of matches per second.

Anyhow, if you can figure out the signing I'll do a review and I'm sure I'll go ahead and merge.

And, thanks and welcome back!

PR #467, opened 2026-01-18
match_set reallocates.

sayrer:
I had Claude Code look at this. Benchmarking showed at least 36% in madvise, which means the code is fast, but there might be reallocations of arrays and/or hashtables.

It came up with a bogus optimization that shared things across test runs, but this one is fair.

### DIFF
--- a/core_matcher.go
+++ b/core_matcher.go
@@ -178,7 +178,9 @@ func (m *coreMatcher) matchesForFields(fields []Field, bufs *nfaBuffers) ([]X, e
 	} else {
 		sort.Sort(fieldsList(fields))
 	}
-	matches := newMatchSet()
+	// Reuse the matchSet from buffers to reduce allocations
+	matches := bufs.matches
+	matches.reset()
 	cmFields := m.fields()
 
 	// for each of the fields, we'll try to match the automaton start state to that field - the tryToMatch
--- a/match_set.go
+++ b/match_set.go
@@ -10,6 +10,11 @@ func newMatchSet() *matchSet {
 	return &matchSet{set: make(map[X]bool)}
 }
 
+// reset clears the matchSet for reuse, preserving the allocated map capacity.
+func (m *matchSet) reset() {
+	clear(m.set)
+}
+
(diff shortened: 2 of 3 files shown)


timbray:
Not sure what to do about this.  Nice and clean, passes all the tests, but the effect on performance is… small and mixed.  I have a little benchmark (attached, really must check it in) that I set a lot of store by in terms of matching real-world experience.

Here are 3 runs, before and after applying this change.

Before:
ANYTHING-BUT: 0.369
ANYTHING-BUT: 0.369
ANYTHING-BUT: 0.371
Avg: 2.444
Avg: 2.572
Avg: 2.587
EQUALS_IGNORE-CASE: 1.578
EQUALS_IGNORE-CASE: 1.590
EQUALS_IGNORE-CASE: 1.602
EXACT: 2.598
EXACT: 2.697
EXACT: 2.804
REGEXP: 1.555
REGEXP: 1.555
REGEXP: 1.639
SHELLSTYLE: 1.232
SHELLSTYLE: 1.420
SHELLSTYLE: 1.430

After:
ANYTHING-BUT: 0.363
ANYTHING-BUT: 0.367
ANYTHING-BUT: 0.369
Avg: 2.571
Avg: 2.577
Avg: 2.598
EQUALS_IGNORE-CASE: 1.555
EQUALS_IGNORE-CASE: 1.567
EQUALS_IGNORE-CASE: 1.578
EXACT: 2.804
EXACT: 2.804
EXACT: 2.804
REGEXP: 1.567
REGEXP: 1.614
REGEXP: 1.639
SHELLSTYLE: 1.349
SHELLSTYLE: 1.430
SHELLSTYLE: 1.430

(the units are millions of matches/sec)

Interesting:  
1. the EXACT match performance is (slightly) better and (distinctly) more stable
2. The others are effectively the same, within the limits of random run-to-run variation
3. This is weird, because the flattener takes a bit over 50% of elapsed time last time I checked and making it faster should have a bigger difference. 

Let me sleep on it but I think I'll probably take it; the complexity is not excessive and the EXACT match is the most common use case,.

PR #470, opened 2026-01-20
More allocation avoidance.

sayrer:
I think this patch is valid, and `perf.sh` is doing too much measuring of loading test data. This one will not be an easy rubber-stamp like the other Claude patches. The `perf.sh` numbers don't look like an improvement, but the profiles are much different.

Before:

```
EXACT: 3.088
EXACT: 3.088
EXACT: 3.088
ANYTHING-BUT: 0.533
ANYTHING-BUT: 0.530
ANYTHING-BUT: 0.534
SHELLSTYLE: 1.393
SHELLSTYLE: 1.402
SHELLSTYLE: 1.402
EQUALS_IGNORE-CASE: 1.746
EQUALS_IGNORE-CASE: 1.746
EQUALS_IGNORE-CASE: 1.746
REGEXP: 1.705
REGEXP: 1.705
REGEXP: 1.590
Avg: 2.821
Avg: 2.824
Avg: 2.787
```

After:

```
EXACT: 3.133
EXACT: 3.133
EXACT: 3.133
ANYTHING-BUT: 0.531
ANYTHING-BUT: 0.535
ANYTHING-BUT: 0.534
SHELLSTYLE: 1.323
SHELLSTYLE: 1.440
SHELLSTYLE: 1.459
EQUALS_IGNORE-CASE: 1.746
EQUALS_IGNORE-CASE: 1.761
EQUALS_IGNORE-CASE: 1.746
REGEXP: 1.590
REGEXP: 1.732
REGEXP: 1.761
Avg: 2.822
Avg: 2.820
Avg: 2.878
```

Comparing the original profile to the final one after optimizations, here is the shift in where CPU time is spent:

### Original Profile (High Churn)
The profile was dominated by the Go runtime managing memory (allocating, clearing, sweeping), rather than doing useful work.

```
      flat  flat%   sum%        cum   cum%
     3.25s 14.95% 14.95%      3.25s 14.95%  runtime.madvise                <-- MEMORY
     2.80s 12.88% 27.83%      2.80s 12.88%  runtime.memclrNoHeapPointers   <-- MEMORY
     1.37s  6.30% 34.13%      1.77s  8.14%  runtime.mallocgcTiny           <-- MEMORY
     1.18s  5.43% 39.56%      1.18s  5.43%  runtime.(*gcBits).bytep
     ...
     0.34s  1.56% 68.22%      7.24s 33.30%  quamina...(*fieldMatcher).transitionOn  <-- Actual Work
```
**Interpretation**: Nearly **35%** of the "flat" CPU time was spent purely on memory mechanics (`madvise`, `memclr`, `malloc`). The actual matching logic (`transitionOn`) was buried.

---

### Optimized Profile (Efficient Work)
Now, the profile looks much healthier. The top consumers are dominated by test data loading (which is unavoidable in this benchmark) and actual application logic.

```
      flat  flat%   sum%        cum   cum%
     0.84s 53.85% 53.85%      0.84s 53.85%  runtime.madvise
     0.24s 15.38% 69.23%      0.24s 15.38%  syscall.syscall
     0.09s  5.77% 75.00%      0.09s  5.77%  runtime.memclrNoHeapPointers
     ...
     0.04s  2.56% 82.69%      0.14s  8.97%  quamina...(*flattenJSON).readObject     <-- Actual Work
     ...
     0.01s  0.64% 94.23%      0.26s 16.67%  quamina...(*Quamina).MatchesForEvent    <-- Actual Work
```

**Interpretation**:
1.  **Reduced Overhead**: While `madvise` is still present (likely due to the massive initial data load of `citylots.json` into memory), the raw time spent there dropped from **3.25s** to **0.84s**.
2.  **Logic Visibility**: The Quamina functions like `readObject` and `MatchesForEvent` are now much more prominent relative to the total runtime (excluding the test setup).
3.  **Faster Execution**: The total execution time for the benchmark dropped significantly (from ~17s to ~2s in the `go test` output), meaning the CPU is doing far less "administrative" work for every event matched.

The allocation-heavy functions like `mallocgcTiny` have effectively disappeared from the top of the profile, confirming that the pooling and slab allocation strategies successfully eliminated the allocation churn.

### DIFF
--- a/epsi_closure.go
+++ b/epsi_closure.go
@@ -1,10 +1,7 @@
 package quamina
 
 type epsilonClosure struct {
-	closures  map[*faState][]*faState
-	hotCache  [8]*faState   // Hot states (keys)
-	hotValues [8][]*faState // Corresponding closures (values)
-	hotIndex  int           // Round-robin index for cache replacement
+	closures map[*faState][]*faState
 }
 
 func newEpsilonClosure() *epsilonClosure {
@@ -15,51 +12,15 @@ func newEpsilonClosure() *epsilonClosure {
 
 func (ec *epsilonClosure) reset() {
 	clear(ec.closures)
-	// Clear hot cache
-	for i := range ec.hotCache {
-		ec.hotCache[i] = nil
-		ec.hotValues[i] = nil
-	}
-	ec.hotIndex = 0
 }
 
 func (ec *epsilonClosure) getClosure(state *faState) []*faState {
-	// Check hot cache first (unrolled for performance)
-	if ec.hotCache[0] == state {
-		return ec.hotValues[0]
-	}
-	if ec.hotCache[1] == state {
-		return ec.hotValues[1]
-	}
-	if ec.hotCache[2] == state {
-		return ec.hotValues[2]
-	}
-	if ec.hotCache[3] == state {
-		return ec.hotValues[3]
-	}
-	if ec.hotCache[4] == state {
-		return ec.hotValues[4]
-	}
-	if ec.hotCache[5] == state {
-		return ec.hotValues[5]
-	}
-	if ec.hotCache[6] == state {
-		return ec.hotValues[6]
-	}
-	if ec.hotCache[7] == state {
-		return ec.hotValues[7]
-	}
-
-	// Check main map cache
+	// Check map cache
 	var closure []*faState
 	var ok bool
 	if ec.closures != nil {
 		closure, ok = ec.closures[state]
 		if ok {
-			// Promote to hot cache
-			ec.hotCache[ec.hotIndex] = state
-			ec.hotValues[ec.hotIndex] = closure
-			ec.hotIndex = (ec.hotIndex + 1) & 7 // Wrap around using mask
 			return closure
 		}
 	}
@@ -83,10 +44,6 @@ func (ec *epsilonClosure) getClosure(state *faState) []*faState {
 	}
 	if ec.closures != nil {
 		ec.closures[state] = closure
-		// Add to hot cache
-		ec.hotCache[ec.hotIndex] = state
-		ec.hotValues[ec.hotIndex] = closure
-		ec.hotIndex = (ec.hotIndex + 1) & 7 // Wrap around using mask
 	}
 	return closure
 }


sayrer:
Maybe a separate PR with that benchmark program first?

timbray:
Will have a look at that memory profile later today.  Might add a CPU profile too.

timbray:
Huh, can't figure out the incantation to make the memory profiler ignore the getCl2Lines and just profile the MatchesForEvent & descendants.

The cpu profiler reveals … no surprises.  This particular benchmark really stresses FlattenJSON because the field values are big and the benchmarks use most of them.  I'm going to have to buckle down and invest some time into building an actually useful benchmarker that lets the profiler cast light on what's going on in Quamina's own code.

timbray:
Yeah, it's complicated. The pattern of memory usage varies a lot depending on what kind of patterns you're looking for and what kind of records you're scanning.  And is often going to be dominated by the flattener code.  The patterns that can run in DFA mode (everything but "wildcard" and "regexp") are really hard to improve at this point, they've been profiled and optimized a lot.  The ones that need NFA mode have both automaton-creation and automaton-traversal ugliness, and haven't been really well profiled yet.  Once again, the biggest win would be eliminating epsilon transitions, if Claude could do that I'd be super impressed.

sayrer:
I went at this for a little bit, but the result was that the tests measured data loading more than Quamina. So, looked at a few different angles, and found that every time.

timbray:
> So, looked at a few different angles, and found that every time.

Not sure what you're saying here.  Does this latest version provide a significant performance increase?

sayrer:
The results I consistently found were that these were faster, but that the benchmarks measure too much non-Quamina code. As in, code that Quamina doesn't even call into. Just benchmark setup, data loading, etc.

sayrer:
I'll try it when I get chance (probably in a day or so). I think the benchmark in #473 looks good but needs to have the input randomized in a loop to insert at least CJK characters and emoji. (this is because you get into the really gnarly parts of the unicode lookup tables, and it's easy to optimize for ASCII if you don't do this)

PR #482, opened 2026-01-27
kaizen: Precomputed epsilon closures

sayrer:
Epsilon closures are a property of the automaton structure, not the input data. Once a pattern is added and the NFA is built, the epsilon closure for any given state is fixed and never changes. (see #470)

**Before:** Every time we matched an event, `traverseNFA` called `getClosure(state)` which:
1. Did a map lookup to check if we'd seen this state before
2. If not, traversed epsilon transitions recursively
3. Allocated a new `[]*faState` slice for the result
4. Cached it in a map for future lookups

This happened on every state, every byte, every match.

**After:** When a pattern is added, we walk the automaton once and store each state's closure directly on the `faState` struct:
```go
type faState struct {
    table            *smallTable
    fieldTransitions []*fieldMatcher
    isSpinner        bool
    epsilonClosure   []*faState  // precomputed
}
```

At match time, `traverseNFA` just reads `state.epsilonClosure` - a direct slice access instead of a map lookup.

**Tradeoff:** Small increase in build-time cost and memory per state, but eliminates map lookups and allocations from the hot path. 

1. **nfa.go**:
   - Added `epsilonClosure []*faState` field to `faState` struct
   - Added `precomputeEpsilonClosures()`, `precomputeClosuresRecursive()`, `computeClosureForState()`, and `traverseEpsilonsForClosure()` functions
   - Updated `traverseNFA` to use precomputed closures with fallback for tests

2. **value_matcher.go**:
   - Call `precomputeEpsilonClosures()` after setting `startTable`, but only when `isNondeterministic` is true

~~The `eClosure` field in `nfaBuffers` was kept for backward compatibility with tests that call `traverseNFA` directly without going through the normal `addPattern` path.~~

### DIFF
--- a/regexp_nfa_test.go
+++ b/regexp_nfa_test.go
@@ -54,9 +54,10 @@ func TestRegexpWorkbench(t *testing.T) {
 	// ~p{L}~p{Zs}~p{Nd}
 	// ((ab){2})?
 	// ([0-9]+(~.[0-9]+){3})
+	// (ab){2,}
 	pp := newPrettyPrinter(2355)
-	matches := applyAndRunRegexp(t, "(ab){2,}", "ababab", pp)
-	if matches != 1 {
+	matches := applyAndRunRegexp(t, "a?", "b", pp)
+	if matches != 0 {
 		t.Error("Workbench")
 	}
 }
@@ -65,7 +66,7 @@ func applyAndRunRegexp(t *testing.T, regexp string, match string, pp printer) in
 	t.Helper()
 	qm := []byte(`"` + match + `"`)
 	fa := faFromRegexp(t, regexp, pp)
-	// fmt.Println("FA:\n" + pp.printNFA(fa))
+	fmt.Println("N:\n" + pp.printNFA(fa))
 	var transitions []*fieldMatcher
 	bufs := newNfaBuffers()
 	matches := traverseNFA(fa, qm, transitions, bufs, pp)
@@ -310,16 +311,16 @@ func TestMakeDotRegexpNFA(t *testing.T) {
 }
 
 func TestAddRuneTreeEntry(t *testing.T) {
-	var root runeTreeNode = make([]*runeTreeEntry, byteCeiling)
+	root := &skinnyRuneTreeNode{}
 	bbs := [][]rune{
 		{'a', 'b', 'c'},
 	}
 	dest := &faState{}
 	for _, runes := range bbs {
 		for _, r := range runes {
-			addRuneTreeEntry(root, r, dest)
+			addSkinnyRuneTreeEntry(root, r, dest)
 		}
-		fmt.Printf("RL: %d\n", len(root))
+		fmt.Printf("RL: %d\n", len(root.byteVals))
 	}
 }
 
--- a/rune_range.go
+++ b/rune_range.go
@@ -116,13 +116,13 @@ func makeAndCacheRuneRangeFA(rr RuneRange, next *faState, name string, pp printe
 		panic("Invalid rune range")
 	}
 
-	var root runeTreeNode = make([]*runeTreeEntry, byteCeiling)
+	root := &skinnyRuneTreeNode{}
 
 	// for each rune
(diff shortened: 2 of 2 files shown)


nfa.go
```diff
@@ -289,6 +304,68 @@ func traverseNFA(table *smallTable, val []byte, transitions []*fieldMatcher, buf
 	return newTransitions.all()
 }
 
+// precomputeEpsilonClosures walks the automaton starting from the given table
+// and precomputes the epsilon closure for every reachable faState.
```
timbray: Curious... wouldn't epsilonClosure.getClosure() do all this?  I must be missing something.

sayrer: I'll look. It's probably just from hammering on the profile too much (there are two paths here, but if the test-only one goes away, it will be clearer)

timbray: Let me know when you want me to have a look.

timbray:
I guess the assumption has to be that we're going to go ahead and compute the closure for every pattern that needs an NFA: shellstyle, wildcard, and regexp.  But note that not all regexes need NFAs, for example `.`, `[a-z0-9]`, `[^z]`, `(a|b){3}`, `~p{}` so the whole read_regexp and makeRegexpNFA are going to have to be refactored a bit.

Anyhow, obviously we'll take this given the measurements.  Needs a bit of work; I apologize in advance for incoming pedantic grumbles about function naming and so on.

In parallel I'm working on code in a file named memory_cost.go, the contents of which are obvious.

sayrer:
> I guess the assumption has to be that we're going to go ahead and compute the closure for every pattern that needs an NFA: shellstyle, wildcard, and regexp. But note that not all regexes need NFAs, for example `.`, `[a-z0-9]`, `[^z]`, `(a|b){3}`, `~p{}` so the whole read_regexp and makeRegexpNFA are going to have to be refactored a bit.
> 
> Anyhow, obviously we'll take this given the measurements. Needs a bit of work; I apologize in advance for incoming pedantic grumbles about function naming and so on.

Don't worry about that part, but this is an underlying flaw in the naming. `nfa.go` contains `traverseDFA` etc. I'd suggest getting this one and the one you have there in before changing lots of names.

epsi_closure.go
```diff
@@ -1,54 +1,63 @@
 package quamina
 
-type epsilonClosure struct {
-	closures map[*faState][]*faState
+// precomputeEpsilonClosures walks the automaton starting from the given table
+// and precomputes the epsilon closure for every reachable faState.
+func precomputeEpsilonClosures(table *smallTable) {
+	visited := make(map[*smallTable]bool)
```
timbray: why not just computeClosureForNfa(table, make(map[*smallTable]bool))

timbray:
> Don't worry about that part, but this is an underlying flaw in the naming. `nfa.go` contains `traverseDFA` etc. I'd suggest getting this one and the one you have there in before changing lots of names.

Let's clean up incrementally and when we check in names, be satisfied with them.  I agree that nfa.go should be automaton.go or some such.

nfa.go
```diff
@@ -129,22 +122,24 @@ func (nb *nfaBuffers) getTransmap() *transmap {
 }
 
 // nfa2Dfa does what the name says, but as of 2025/12 is not used.
+// Requires that precomputeEpsilonClosures has been called on the NFA.
 func nfa2Dfa(nfaTable *smallTable) *faState {
-	startNfa := []*faState{{table: nfaTable}}
-	ec := newEpsilonClosure()
-	return n2dNode(startNfa, newStateLists(), ec)
+	startState := &faState{table: nfaTable}
+	computeClosureForState(startState)
```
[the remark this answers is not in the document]
sayrer: As written, this was required because startState didn't have its closures computed, even though the wrapped one did. I changed the API a little, so it's a bit cleaner now.

timbray:
Yeah, you're right about the tests. If you look close, a lot of them have this clumsy wrap-the-word-with `"` code.

timbray:
Eh, never mind. I have way worse merge problems.

PR #486, opened 2026-02-01
kaizen: Flatten epsilon targets

sayrer:


timbray:
Hmmm
```
main

Patterns/sec: 1494.5
SmallTables 8405 (splices 998, avg 3.132, max 245, epsilons avg 5.786, max 164) singletons 0
Huge-machine events/sec: 46317.8

Patterns/sec: 1495.7
SmallTables 8405 (splices 998, avg 3.132, max 245, epsilons avg 5.786, max 164) singletons 0
Huge-machine events/sec: 43083.3

flatten-epsilon-targets

Patterns/sec: 829.2
SmallTables 14573 (splices 3, avg 3.395, max 1118, epsilons avg 6.638, max 1117) singletons 0
Huge-machine events/sec: 17777.8

Patterns/sec: 834.0
smalltables 14573 (splices 3, avg 3.395, max 1118, epsilons avg 6.638, max 1117) singletons 0
Huge-machine events/sec: 17391.3
```
Seems to slow things down?  Sorry, will have to reconstruct step-by-step to make sure I don't have it backward or something…

timbray:
Actually I think my results are valid.  In fact I changed both branches to use 2000 patterns.  So yeah, I don't think this helps, but you are definitely looking in the right place.

sayrer:
It is possible to speed up Events/sec at great cost to pattern building and memory footprint. I think this is a bad tradeoff. The current patch has no downsides, so I think that one is better, but it can be done. That means the code is about as fast as it can be without horse trading.

```
Results with direct array lookup:                                                                                                                              
  ┌─────────────────┬───────────┬────────┬────────┐                                                                                                              
  │     Metric      │  Before   │ After  │ Change │                                                                                                              
  ├─────────────────┼───────────┼────────┼────────┤                                                                                                              
  │ Events/sec      │ 40,000    │ 82,868 │ +107%  │                                                                                                              
  ├─────────────────┼───────────┼────────┼────────┤                                                                                                              
  │ Patterns/sec    │ 3,244     │ 830    │ -74%   │                                                                                                              
  ├─────────────────┼───────────┼────────┼────────┤                                                                                                              
  │ Benchmark ns/op │ 33,177    │ 20,139 │ -39%   │                                                                                                              
  ├─────────────────┼───────────┼────────┼────────┤                                                                                                              
  │ Memory/table    │ ~28 bytes │ ~2KB   │ +70x   │                                                                                                              
  └─────────────────┴───────────┴────────┴────────┘
```                                                                                                              
  Summary:                                                                                                                                                       
  - Matching is 2x faster (O(1) vs O(n) lookup)                                                                                                                  
  - Building is 4x slower (computing fastSteps arrays)                                                                                                           
  - Memory usage increased by ~14MB for the automaton

timbray:
Not sure quite where things stand. I agree that speeding up matching with that level of cost to automaton building is not a win.  Does that apply to this PR as a whole or is there a version with better trade-offs that we should consider merging, or what?

sayrer:
OK, that is indeed better, so the review comment was correct.

The basic insight is never trust a map/vector/string allocation without a capacity, and then also question whether there needs to be a new one at all. Not brilliant, but effective.

nfa.go
```diff
@@ -295,6 +295,36 @@ func makeFaStepKey(s1, s2 *faState) faStepKey {
...
 }
 
+// flattenEpsilonTargets collects all non-epsilon-only states reachable via
+// epsilon transitions from state1 and state2. This prevents deep nesting of
+// splice states that would otherwise accumulate during repeated merges.
+func flattenEpsilonTargets(state1, state2 *faState) []*faState {
+	seen := make(map[*faState]bool, 8)
+	targets := make([]*faState, 0, 4)
+
+	var collect func(s *faState)
```
timbray: Ha, once again you & Claude know more Go than I do.  I've never got around to using closures, but this obviously saves a lot of stack burn.

sayrer: I think we might also want to look at a CPU profile of traverseDFA. There are some easy wins in there, I think.

timbray: Anyhow, I have verified that this PR doesn't make nfa2Dfa worse, so will merge.

sayrer: Also, Melinda is smarter than I am, in case that was not obvious.

Issue #481, opened 2026-01-27
NFA => DFA optimization


timbray:
It turns out that traversing NFAs, which are produced by any instance that has wildcard or regexp patterns, is a lot slower than with DFAs.  More than I'd like, I'm going to have to re-write the performance section of README.md. However, the finite-automaton theory says that any NFA can be turned into a DFA, and I just finished debugging and testing the `nfa2Dfa()` function.  It makes a huge difference, tests where the instance has a significant number of NFA valueMatchers, converting everything to a DFA makes things run at least twice as fast and occasionally much, much faster.  But, we can't just go ahead and blast all the NFAs into DFAs, because:

1. It slows down the `addPattern()` call quite a bit.
2. The state machines occupy a lot more memory.
3. In some cases, the slowdown is O(2^N), ouch. An example is in `TestShellStyleBuildTime`, where if you put 1000 patterns in, it basically never comes back.

So… there are things we could do.

### Heuristics

Put on a hard-wired time limit, for example one second, and if any NFA=>DFA conversion runs longer than that, give up and let it be an NFA. This means that users don't have to think about it and will usually get about the best result.

### Controls

Have a new `Option` on the `New()` function to specify one of three modes: Never optimize, Always optimize, or Optimize-with-limit, where the caller provides the time threshold and Quamina gives up on the DFA conversion if it runs longer than that.  You could imagine an even finer-grained control where you could specify this per-Field but we could do that later.

This is good for people who need flexibility i.e. they're doing a lot of `AddPattern()` calls and want low latency on that. Or on the other end of the scale, are willing to amortize a one-time build cost for a big NFA to buy themselves minimal latency down the pipe.

### Stats

We could add a call to report the current stats of the state machine, i.e. the data that is now reported by `matcherStats()`, you can see the text output quite a bit in the unit-test outputs. This would give callers the option of doing a test run and finding out how much DFA conversions are going to bloat the memory consumption of a Quamina instance.

timbray:
Hey @embano1 I like your thinking about APIs, got any good ideas on this one?

sayrer:
Oh, yes. It is a good example of hallucinations. But some of the ideas are good. I liked `func (t *smallTable) complexity()` better than memory fwiw.

timbray:
[...]

PR #492, opened 2026-02-10
Add a benchmark targeting NFA to DFA tradeoffs.

sayrer:
Here's the tradeoff this file attempts to measure (see #481):

**Small state space, eager fits in budget** → eager wins, it's faster, no cache overhead.

**Large state space, predictable input** → lazy wins by a mile. Eager can't even attempt it.

**Large state space, adversarial input** → lazy falls back to NFA-with-overhead. Eager also falls back to NFA because it blew the budget. You're in the same place, maybe ~2x slower from your varied input benchmarks — but that's 2x slower than a path that was already the fallback.


This file contains 5 shellstyle wildcard benchmarks designed to characterize NFA vs DFA tradeoffs:

  1. **BenchmarkShellstyleSimpleWildcard** (line 16) — Simple prefix*suffix patterns like "a*b" where an eager DFA would be ~3 states. Tests whether simple wildcards deserve DFA treatment.                                                                                 
  2. **BenchmarkShellstyleNarrowInput** (line 90) — Wide Unicode patterns (anchors from ASCII, CJK, mixed scripts) but narrow input alphabets (digits, lowercase, etc.). Shows a demand-driven DFA only needs states for bytes actually seen.
  3. **BenchmarkShellstyleWidePatternsScaling** (line 219) — Scales from 8 to 512 patterns with multi-script anchors but ASCII-digit-only input. Isolates how a lazy DFA cache stays small regardless of Unicode coverage.                                                   
  4. **BenchmarkShellstyleSimpleWildcardScaling** (line 300) — Scales from 1 to 26 independent simple patterns to show even modest collections benefit from DFA conversion.
  5. **BenchmarkShellstyleZWJEmoji** (line 373) — Worst case: ZWJ emoji sequences (15-25+ bytes per glyph) mixed with Japanese text. Shared leading bytes (0xE2, 0xE3, 0xE4) force massive NFA branching.

```
  SimpleWildcard — 325-422 ns/op, 1 alloc. Very fast for simple prefix*suffix patterns.
                                                                                                                                                                                                                                                                         
  SimpleWildcardScaling — ~370 ns/op regardless of pattern count (1-26). Scaling is flat, which is good.                                                                                                                                                                 
                                                                                                                                                                                                                                                                         
  NarrowInput — Scales roughly linearly with pattern count. Notable: multi-byte input (narrow CJK) is ~3-4x slower than ASCII digits, showing the cost of UTF-8 byte-level branching in the NFA.                                                                         
                                                                                                                                                                                                                                                                         
  WidePatternsScaling — Shows superlinear scaling: 987 ns (8 patterns) → 148 μs (512 patterns). The jump from 256→512 patterns (31→149 μs, ~4.8x) suggests the NFA traversal cost is growing faster than linearly.                                                       

  ZWJEmoji — 5.7-40 μs with 14-15 allocs/op. The high allocation count (vs 1 alloc for simpler benchmarks) and the per-op cost confirm that ZWJ sequences with shared leading bytes are expensive for NFA traversal.
```

### DIFF
--- a/regex_nfa_dfa_bench_test.go
+++ b/regex_nfa_dfa_bench_test.go
@@ -389,12 +389,17 @@ func BenchmarkShellstyleZWJEmoji(b *testing.B) {
 		"👁\uFE0F\u200D🗨\uFE0F",   // eye in speech bubble
 	}
 
-	// Japanese text that shares leading UTF-8 bytes with ZWJ sequences.
+	// Japanese text using leading UTF-8 bytes near the ZWJ range.
 	// Hiragana (U+3040-309F): 0xE3 0x81 0x80 - 0xE3 0x82 0x9F
 	// Katakana (U+30A0-30FF): 0xE3 0x82 0xA0 - 0xE3 0x83 0xBF
 	// CJK (U+4E00+):          0xE4-0xE9 ...
-	// All start with 0xE3/0xE4+ which the NFA cannot distinguish from
-	// 0xE2 (ZWJ prefix) without reading the second byte.
+	// The ZWJ byte sequence (0xE2 0x80 0x8D) shares its leading byte
+	// 0xE2 with hundreds of other BMP codepoints (U+2000-U+2FFF), so
+	// the NFA cannot distinguish a ZWJ from other 0xE2-prefixed characters
+	// without reading the second and third bytes. Combined with the Japanese
+	// filler (0xE3, 0xE4+) and variation selectors (0xEF), the wildcard's
+	// self-loop must handle dense multi-byte traffic across several leading
+	// byte ranges.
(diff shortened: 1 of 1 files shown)


sayrer:
> Cool benchmarks, thanks, will probably adopt. But I'm missing something, the benchmarks don't call nfa2Dfa so how do you arrive at the conclusions up at the top of this thread?

You can only do damage to these with a lazy or eager (nfa2dfa) implementation. So, these set the baseline with always-NFA in the presence of wildcards. So, if you look at the patches here (picking and choosing), you'll see it: https://github.com/timbray/quamina/compare/main...sayrer:quamina:lazy_dfa

PR #494, opened 2026-02-16
kaizen: fix duplicate states during closure traversal

sayrer:
Here's what dedup_fix does relative to main:  
                                                                           
  Problem: During FA merges (especially with shell-style * patterns),
  the same smallTable can end up referenced by multiple faState
  nodes. When epsilon closures are computed, these duplicate table
  pointers cause the same logical     state to appear multiple times,
  leading to redundant work during NFA traversal.  
   
  Solution — table-pointer dedup in epsilon closure
  (epsi_closure.go):  
                                   
  - Adds atableSeen map[*smallTable]bool to closureBuffers, tracked
    alongside the existing closureSet (which is keyed by *faState).
  - In traverseEpsilons, before adding a non-epsilon-only state to the
    closure, it checks whether that state's *smallTable has already
    been seen. If so, the state is skipped as a duplicate.
  - This eliminates duplicate NFA states at the source — the closure
    itself is smaller, so every downstream consumer benefits.

  Consequence — removal of runtime dedup in traverseNFA (nfa.go):

  - The len(nextStates) > 500 sort+compact block is removed. This was
    a safety valve for "toxically-complex regexps" where epsilon
    loops caused huge nextStates buildups. With duplicates eliminated
    at closure-computation time, this is no longer needed.
  - The cmp and slices imports are dropped accordingly.

  In short: instead of deduplicating at traversal time
  (expensive, per-event), this branch deduplicates at build time
  (once, when patterns are added) by recognizing that distinct
  faState pointers sharing the same smallTable are logically
  identical.

```
    Timing (sec/op)                                                                                                                                                                                                                        
  ┌──────────────────────────────────────────┬─────────┬───────────┬──────────────────────────────────────┐                                                                                                                              
  │                Benchmark                 │  main   │ dedup_fix │                change                │                                                                                                                              
  ├──────────────────────────────────────────┼─────────┼───────────┼──────────────────────────────────────┤                                                                                                                              
  │ CityLots                                 │ 5.878µs │ 5.774µs   │ -1.78% (p=0.002)                     │                                                                                                                              
  ├──────────────────────────────────────────┼─────────┼───────────┼──────────────────────────────────────┤                                                                                                                              
  │ JsonFlattener_LastField                  │ 9.648µs │ 9.428µs   │ -2.27% (p=0.002)                     │
  ├──────────────────────────────────────────┼─────────┼───────────┼──────────────────────────────────────┤
  │ JsonFlattener_Evaluate_LastField         │ 9.777µs │ 9.535µs   │ -2.47% (p=0.002)                     │
  ├──────────────────────────────────────────┼─────────┼───────────┼──────────────────────────────────────┤
  │ JsonFlattener_Evaluate_ContextFields     │ 431.4ns │ 425.6ns   │ -1.33% (p=0.002)                     │
  ├──────────────────────────────────────────┼─────────┼───────────┼──────────────────────────────────────┤
  │ JsonFlattener_Evaluate_MiddleNestedField │ 9.154µs │ 9.029µs   │ -1.37% (p=0.002)                     │
  ├──────────────────────────────────────────┼─────────┼───────────┼──────────────────────────────────────┤
  │ ShellstyleMultiMatch                     │ 32.94µs │ 32.49µs   │ ~1.4% faster (not stat. significant) │
  ├──────────────────────────────────────────┼─────────┼───────────┼──────────────────────────────────────┤
  │ Others                                   │         │           │ no significant change                │
  └──────────────────────────────────────────┴─────────┴───────────┴──────────────────────────────────────┘
  
  Memory: CityLots dropped from 15 to 14 B/op (-6.67%). Everything
  else unchanged. Zero allocations across the board (as before).

  Summary: Statistically significant 1.3-2.5% speedups across most
  benchmarks, with the geomean improvement at -0.66%. The gains are
  modest but consistent — the table-pointer dedup is eliminating
  redundant work in epsilon closures. The biggest wins are in the
  flattener benchmarks that exercise deeper automata. No regressions
  anywhere.
```

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go ${{ matrix.go-version }}
-        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
         with:
           go-version: ${{ matrix.go-version }}
         id: go
--- a/.github/workflows/go-lint.yaml
+++ b/.github/workflows/go-lint.yaml
@@ -24,7 +24,7 @@ jobs:
         fetch-depth: 1
 
     - name: Set up Go ${{ matrix.go-version }}
-      uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+      uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
       with:
         go-version: ${{ matrix.go-version }}
       id: go
--- a/.github/workflows/go-unit-tests.yaml
+++ b/.github/workflows/go-unit-tests.yaml
@@ -34,7 +34,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go ${{ matrix.go-version }}
-        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
         with:
           go-version: ${{ matrix.go-version }}
         id: go
--- a/v2_bench_test.go
+++ b/v2_bench_test.go
@@ -92,16 +92,12 @@ func BenchmarkShellStyleBuildTime(b *testing.B) {
 	}
 
 	q, _ := New()
-	before := time.Now()
 	for i := range words {
 		err := q.AddPattern(starWords[i], patterns[i])
 		if err != nil {
 			b.Fatal("AddP: " + err.Error())
 		}
 	}
-	elapsed := time.Since(before).Seconds()
-	b.Logf("Patterns/sec: %.1f", float64(len(words))/elapsed)
-	b.Log(matcherStats(q.matcher.(*coreMatcher)))
 
 	// Build events: original words and expanded words
 	type event struct {
@@ -129,9 +125,6 @@ func BenchmarkShellStyleBuildTime(b *testing.B) {
 			}
 		}
 	}
-	elapsed = float64(b.Elapsed().Seconds())
-	count := float64(b.N)
-	b.Logf("%.0f events/sec", count*float64(len(events))/elapsed)
 }
 
 // BenchmarkTablePointerDedup benchmarks matching speed for workloads where


sayrer:
> Have the benchmark output been updated following on all those additional pushes? Because they're kind of disappointing as written.

Yeah, they aren't good. But they do show we're in the right department, and there is a new test at least. I was looking in the context of DFA https://github.com/sayrer/quamina/tree/lazy_dfa.

sayrer:
Now we're getting somewhere with this test I added called `TestEpsilonClosureSizes`:

```
  ┌──────────────────────────┬────────────────────────┬─────────────────────────┬────────────┬─────────────┬────────────────────┬─────────────────────┬─────────┐
  │         Workload         │ Closure Entries (main) │ Closure Entries (dedup) │ Max (main) │ Max (dedup) │ Matches/sec (main) │ Matches/sec (dedup) │ Speedup │
  ├──────────────────────────┼────────────────────────┼─────────────────────────┼────────────┼─────────────┼────────────────────┼─────────────────────┼─────────┤
  │ 6-regexp + 12-shell      │ 4,392                  │ 4,371                   │ 19         │ 13          │ 240K               │ 380K                │ 1.6x    │
  ├──────────────────────────┼────────────────────────┼─────────────────────────┼────────────┼─────────────┼────────────────────┼─────────────────────┼─────────┤
  │ 20 nested regexps        │ 334                    │ 261                     │ 60         │ 40          │ 1.1M               │ 3.4M                │ 3.1x    │
  ├──────────────────────────┼────────────────────────┼─────────────────────────┼────────────┼─────────────┼────────────────────┼─────────────────────┼─────────┤
  │ deeply nested            │ 342                    │ 220                     │ 46         │ 26          │ 681K               │ 3.2M                │ 4.7x    │
  ├──────────────────────────┼────────────────────────┼─────────────────────────┼────────────┼─────────────┼────────────────────┼─────────────────────┼─────────┤
  │ overlapping char classes │ 240                    │ 156                     │ 48         │ 24          │ 1.6M               │ 4.5M                │ 2.9x    │
  ├──────────────────────────┼────────────────────────┼─────────────────────────┼────────────┼─────────────┼────────────────────┼─────────────────────┼─────────┤
  │ shell + deep overlap     │ 3,488                  │ 3,410                   │ 37         │ 21          │ 204K               │ 297K                │ 1.5x    │
  └──────────────────────────┴────────────────────────┴─────────────────────────┴────────────┴─────────────┴────────────────────┴─────────────────────┴─────────┘

  The deeply-nested workload ((((a?)*b?)*c?)* patterns) shows the biggest win:
  max closure drops from 46→26, closure entries from 342→220 (36% fewer), and
  matching is 4.7x faster. The overlapping-char-classes workload((([abc]?)*)+,(
  ([bcd]?)*)+, etc.) is nearly 3x faster with max closure cut in half (48→24).

  The pattern that triggers dedup_fix most strongly is nested quantifiers over
  overlapping character classes — (([abc]?)*)+ merged with (([bcd]?)*)+ creates
  many epsilon paths converging on shared tables.
```

sayrer:
```
  ┌──────────────────────────┬─────────┬───────────┬──────────┐                                                                                                                                                   
  │        Benchmark         │  main   │ dedup_fix │  Change  │                                                                                                                                                   
  ├──────────────────────────┼─────────┼───────────┼──────────┤                                                                                                                                                   
  │ 8259Example              │ 5.710µs │ 5.663µs   │ ~ (flat) │                                                                                                                                                   
  ├──────────────────────────┼─────────┼───────────┼──────────┤                                                                                                                                                   
  │ ShellStyleBuildTime      │ 47.74ms │ 48.60ms   │ ~ (flat) │                                                                                                                                                   
  ├──────────────────────────┼─────────┼───────────┼──────────┤                                                                                                                                                   
  │ 6-regexps-12-shell       │ 8.885µs │ 7.816µs   │ -12.04%  │                                                                                                                                                   
  ├──────────────────────────┼─────────┼───────────┼──────────┤
  │ 20-nested-regexps        │ 1,989ns │ 882ns     │ -55.63%  │
  ├──────────────────────────┼─────────┼───────────┼──────────┤
  │ deeply-nested            │ 3,332ns │ 929ns     │ -72.13%  │
  ├──────────────────────────┼─────────┼───────────┼──────────┤
  │ overlapping-char-classes │ 1,489ns │ 664ns     │ -55.39%  │
  ├──────────────────────────┼─────────┼───────────┼──────────┤
  │ shell+deep-overlap       │ 12.39µs │ 10.06µs   │ -18.87%  │
  ├──────────────────────────┼─────────┼───────────┼──────────┤
```


The branch makes one functional change and several cleanup/test changes:                                                                                                                                        
                                   
  **Functional change:** table-pointer dedup in epsilon closure (build-time)                                                                                                                                          
                                                                                                                                                                                                                  
  In `epsi_closure.go`, when computing the epsilon closure of an NFA state, the code now tracks which *smallTable pointers it has already seen via a tableRep map. When multiple faState nodes reachable through    
  epsilon transitions share the same *smallTable, only one representative is kept in the closure — the others are skipped because their byte transitions are identical. A safety check preserves both states if
  they have different fieldTransitions (so no matches are lost).

  This matters for patterns like (([abc]?)*)+ where nested quantifiers create many epsilon paths that converge on the same underlying table. Without dedup, closures blow up with redundant entries, and
  traverseNFA calls step() on the same table repeatedly for each byte. With dedup, closure sizes drop dramatically (e.g. from ~50 entries to ~30 for deeply-nested patterns), and matching is 12-72% faster on
  affected workloads.

  **Removed:** runtime sort+compact dedup in traverseNFA

  Main had a fallback in traverseNFA that sorted and compacted nextStates when the list exceeded 500 entries. This is no longer needed — the build-time closure dedup prevents the blowup at the source. The
  unused stateSet field and getStateSet() method were also removed.

  **Removed:** unused printer parameter from traverseNFA

  The _ printer parameter was dead code. Removed from the signature and all call sites (value_matcher.go, test files).

  **Benchmark fixes**

  fmt.Printf → b.Logf in v2_bench_test.go so benchmark output parses cleanly with benchstat.

  **New tests and benchmarks**

  - TestTablePointerDedup — correctness test with 5 workloads verifying closure stats and match counts
  - TestTablePointerDedupPreservesFieldTransitions — edge case: two states sharing a table but with different field transitions are both preserved
  - BenchmarkTablePointerDedup — 5 sub-benchmarks targeting the dedup sweet spot
  - collectClosureStats helper for walking NFA structure

timbray:
``` git checkout main ; go test -run=TestShellStyleBuildTime ; go test -run=TestShellStyleBuildTime ; go test -run=TestShellStyleBuildTime ; git checkout dedup_fix ; go test -run=TestShellStyleBuildTime ; go test -run=TestShellStyleBuildTime ; go test -run=TestShellStyleBuildTime```

File: small_table.go [hunk 1]
```diff
@@ -41,7 +41,9 @@ type smallTable struct {
 	ceilings       []byte
 	steps          []*faState
 	epsilons       []*faState
-	lastVisitedGen uint64 // generation counter for epsilon closure traversal
+	lastVisitedGen uint64   // generation counter for epsilon closure traversal
```
timbray: Do you really need two distinct generation counters?  In my work on memory_cost (slow, been distracted) which also needs a generation counter, it looked to me I could re-use the existing one.

timbray (commented):
This adds noticeable complexity and increases the size of smallTable by either 16 bytes (if closureRep == nil) or 40 bytes (closureRep != nil). Given that the kind of scenarios we're addressing are those with tens of thousands of smallTables, this is significant.

There have been a lot of updates to this PR, I lost track a bit of where the benefit occurs, aside from losing the ugliness of the sort/collect-on-500. Could you summarize the win?

epsi_closure.go
```diff
@@ -4,6 +4,7 @@ package quamina
 // Each smallTable stores the generation it was last visited in, avoiding
 // the need for a visited map.
 var closureGeneration uint64
+var closureRepGeneration uint64
```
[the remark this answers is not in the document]
sayrer: "representative", I left a longer explanation below.

timbray:
OK, will give it a close review and probably take it.  But could you check one thing? Do you really need two generation fields in smallTable?  I can guarantee you don't need more than one for build-time operations because those are guaranteed single-threaded.

sayrer:
> OK, will give it a close review and probably take it. But could you check one thing? Do you really need two generation fields in smallTable? I can guarantee you don't need more than one for build-time operations because those are guaranteed single-threaded.

Facts. We can jam this into less memory with no perf loss.

```
  ┌────────────────────────────────────────────┬─────────┬─────────┬───────────────┐
  │                 Benchmark                  │  main   │ branch  │  improvement  │
  ├────────────────────────────────────────────┼─────────┼─────────┼───────────────┤
  │ 8259Example                                │ 5.683µs │ 5.676µs │ ~ (no change) │
  ├────────────────────────────────────────────┼─────────┼─────────┼───────────────┤
  │ ShellStyleBuildTime                        │ 48.02ms │ 48.21ms │ ~ (no change) │
  ├────────────────────────────────────────────┼─────────┼─────────┼───────────────┤
  │ TablePointerDedup/6-regexps-12-shell       │ 8.804µs │ 7.747µs │ -12%          │
  ├────────────────────────────────────────────┼─────────┼─────────┼───────────────┤
  │ TablePointerDedup/20-nested-regexps        │ 1968ns  │ 893ns   │ -55%          │
  ├────────────────────────────────────────────┼─────────┼─────────┼───────────────┤
  │ TablePointerDedup/deeply-nested            │ 3322ns  │ 934ns   │ -72%          │
  ├────────────────────────────────────────────┼─────────┼─────────┼───────────────┤
  │ TablePointerDedup/overlapping-char-classes │ 1500ns  │ 666ns   │ -56%          │
  ├────────────────────────────────────────────┼─────────┼─────────┼───────────────┤
  │ TablePointerDedup/shell+deep-overlap       │ 12.43µs │ 10.04µs │ -19%          │
  └────────────────────────────────────────────┴─────────┴─────────┴───────────────┘

  Zero regressions on existing benchmarks, 12-72% faster on the dedup-heavy workloads.
   Geomean -37%.
```

epsi_closure.go
```diff
@@ -58,28 +60,61 @@ func closureForStateWithBufs(state *faState, bufs *closureBuffers) {
...
-	closure := make([]*faState, 0, len(bufs.closureSet))
-	for s := range bufs.closureSet {
+	// Table-pointer dedup: when multiple states in the closure share the
+	// same *smallTable, their byte transitions are identical, so only one
+	// representative is needed. This is done as a post-pass over the
+	// closure list rather than during traversal to keep traverseEpsilons
+	// zero-overhead. States with different fieldTransitions are preserved.
+	closureGeneration++
+	closure := make([]*faState, 0, len(bufs.closureList))
+	for _, s := range bufs.closureList {
```
timbray: we normally use `state` as the variable name for a *faState

sayrer:
Hmm, this is perplexing. The NFA version is so fast now that the DFA is not worth doing. I would recommend taking this patch with its 2% regression on matching (that is now twice as fast) and then we can inspect more.

sayrer:
OK, yeah. The DFA version can wing this one, so it's not worth debating 2% on the already really fast NFA version.

```
  ┌───────────────────────────────┬─────────┬──────────┬─────────────┐ 
  │           Benchmark           │  main   │ lazy_dfa │   Change    │  
  ├───────────────────────────────┼─────────┼──────────┼─────────────┤ 
  │ ShellstyleMultiMatch          │ 31.80µs │ 10.29µs  │ -67.65%     │  
  ├───────────────────────────────┼─────────┼──────────┼─────────────┤ 
  │ CityLots                      │ 5.862µs │ 5.761µs  │ -1.72%      │
  ├───────────────────────────────┼─────────┼──────────┼─────────────┤
  │ NumberMatching                │ 881.2ns │ 853.1ns  │ -3.18%      │ 
  │ JsonFlattener (eval, context) │ 426.6ns │ 407.2ns  │ -4.55%      │
  ├───────────────────────────────┼─────────┼──────────┼─────────────┤
  │ JsonFlattener (eval, last)    │ 9.845µs │ 9.682µs  │ -1.65%      │
  ├───────────────────────────────┼─────────┼──────────┼─────────────┤
  │ JsonFlattener (eval, middle)  │ 9.158µs │ 9.059µs  │ -1.08%      │
  ├───────────────────────────────┼─────────┼──────────┼─────────────┤
  │ 8259Example                   │ 5.695µs │ 4.813µs  │ ~-15% (n=1) │
  └───────────────────────────────┴─────────┴──────────┴─────────────┘
  

  The lazy DFA branch shows a ~68% speedup on ShellstyleMultiMatch (the primary
  target), with small improvements across other benchmarks and 0 allocs/op on
  all hot paths. No regressions.

```

sayrer:
`go test -bench=BenchmarkShellstyleMultiMatch -benchmem -count=6 -run='^$'  `

sayrer:
But none of this work is in this patch. It's only to point out that we should not sweat 2%, since we can smash it.

sayrer:
> Anyhow, sorry

I don't understand Canadian :)

> building DFAs once per thread is horribly wasteful of memory. In my experience with Quamina's predecessor, basically all the production deployments were multi-threaded.

Oh, that's in there. I do a lot of sloppy ones to see if an idea is worth exploring. I have the memory footprint one already, but it needs the good NFA and eager DFA first.

You're right to insist on clean patches, but I check them out with messy ones before polishing a turd.

small_table.go
```diff
@@ -42,6 +42,16 @@ type smallTable struct {
 	steps          []*faState
 	epsilons       []*faState
 	lastVisitedGen uint64 // generation counter for epsilon closure traversal
+	// closureRepGen records which closureGeneration this table's
+	// representative was set in. If it equals the current global
+	// closureGeneration, then closureRep is valid; otherwise, the
+	// table has not yet been seen in this dedup pass.
+	closureRepGen uint64
```
timbray: I suggest that closureGen is nicely symmetrical with lastVisitedGen, not sure Rep belongs on the name

sayrer:
OK, these all seem reasonable. I'll get them in the morning.

For the lazy DFA design, you can get it to be a global cache, and this is better in some ways. Once one goroutine discovers an optimization, everyone gets it. I hacked it out in that branch I would not submit as a PR. I think there would need to be an intermediate stage where the lazy DFA is wastefully per-goroutine (as you point out). But this is just an investigation: can you get it faster by doing this? I don't know for sure because the NFA is getting better, but I suspect so. Once that is proven, then there's a tricky global cache management problem. I think it's not possible to aim for the end state right away, since it will be unreviewable. But that's where my head is.

PR #516, opened 2026-04-15
api: Add Get/Set Memory Budget

timbray:
I haven't actually enabled the nfa2Dfa code yet, want to first consider the memory_cost logic.

Check out the new memory_bench_test, the numbers are kind of horrifying but it shows the memory_cost machinery working.


go.sum
```diff
@@ -0,0 +1,2 @@
+quamina.net/go/quamina v1.5.1 h1:hrJOsy6wGieU4Jae+Ei4U9Rwic579jEE5+OKkg7gs6E=
```
timbray: Why is this here? How do I make it go away?

timbray: Ok, figured it out, will. push fix before merging

timbray:
image
image
image

PR #520, opened 2026-04-16
kaizen: core_matcher: pool nfaBuffers and flattener in the test helper path

sayrer:
This one isn't a real win, but it's getting annoying to have the benchmarks with different behavior than the production path. When some flavor of more aggressive DFA optimization happens, this will be good.

matchesForJSONWithFlattener previously allocated a fresh *nfaBuffers per call, and matchesForJSONEvent additionally allocated a fresh *flattenJSON per call. These helpers are the path used by most tests and benchmarks (the production *Quamina.MatchesForEvent already reuses both).

Add two package-level sync.Pools and Get/Put on each call. On Put, re-seat bufs.resultBuf with a fresh slice so the []X returned to the caller can't be written through by the next pool user — the old backing array stays exclusive to the caller.

Bench (Apple M1 Ultra, n=6):

  NumberMatching-20   890.2n ->  498.3n   -44.03%
                      2288 B  ->  280 B   -87.76%
                      10 allocs -> 3 allocs
  geomean (10 benchmarks):              -4.71% time, -18.39% B/op

Other benchmarks that route through *Quamina.MatchesForEvent rather than the helpers are unchanged in behavior; a few show +1-3% noise likely from code layout (they don't touch the pooled path).

Tested #519 and #520 together under the race detector and it was clean, but it took 117 seconds.


timbray (approved):

timbray (commented):
Currently failing memory_cost_test, will sit on this till we get that stuff sorted out.

timbray (approved):

PR #527, opened 2026-05-18
Regex nfa dfa benchmarks

sayrer:
These are tests that subsume #492.


regex_nfa_dfa_bench_test.go
```diff
@@ -0,0 +1,498 @@
...
+		{"a*b", "a*b"},
+		{"foo*bar", "foo*bar"},
+		{"x*y*z", "x*y*z"},
+		{"he*lo", "he*lo"},
+	}
+
+	for _, sp := range simplePatterns {
+		b.Run(sp.name, func(b *testing.B) {
+			q, _ := New()
+			pattern := fmt.Sprintf(`{"val": [{"shellstyle": %q}]}`, sp.shellstyle)
```
timbray, line 31: Not a blocker but let's use wildcard rather than shellstyle in the future.

timbray (approved):

regex_nfa_dfa_bench_test.go
sayrer: Sorry to be an epic nitpicker, but there is a problem here. "shellstyle" is the description of the overall pattern, while "wildcard" is the specific production. I don't care how this distinction is resolved, but it's there.

timbray:
On May 29, 2026 at 6:26:02 PM, RS ***@***.***> wrote:

> Sorry to be an epic nitpicker, but there is a problem here. "shellstyle"
> is the description of the overall pattern, while "wildcard" is the specific
> production. I don't care how this distinction is resolved, but it's there.
>
I yield to no-one in the pickiness of my nits.  What happened was, I
implemented shellstyle but stupidly forgot to put in escaping for *, but
then couldn’t change the API, so wildcard is just shellstyle with escaping
for * and then of course \.  -T

sayrer:
I am surprised by your strong opinion here, although I do not disagree with it. I'll try to knock out something to cover this issue tomorrow morning. Happy to do it your way, it's just confusing.

PR #535, opened 2026-05-31
kaizen: Embed smalltable

sayrer:
This one carries #521 further, fixes some bugs that caused buildshellstyle quadratic behavior, and also speeds up Nfa2dfa.


timbray:
(afk for a while, will take a closer look)

timbray:
OK, checked out 535 and ran the research-main CSV generator. I'll attach the xlsx files and a graph of sec/100-addP-calls

may30
pr535
[2026-05-31.xlsx](https://github.com/user-attachments/files/28443966/2026-05-31.xlsx)
[2026-05-30.xlsx](https://github.com/user-attachments/files/28443967/2026-05-30.xlsx)

timbray:
Now, granted, all these AddPattern calls are wildcards, and the epsilon merging gets kind of ugly, so probably not a representative workload.

timbray:
I'm just trying to figure out whether I care about this.  Don't have an opinion yet.

sayrer:
I just got that workload, seems like I can do some damage to it with some small changes.

timbray:
I gotta say, if those numbers are correct, the permanent memory saving is just not that dramatic and the AddPattern impact is pretty significant.  So maybe the thing to do is back out the last PR and then cherrypick the good stuff from the faState/smallTable branch with the add-ons that made nfa2Dfa faster and so on?

sayrer:
I think I can get it with no regression, just measuring.

sayrer:
Even just the first step went well. The closure walk was improved, and had been going over the more space-efficient structures too much.

That's here:
https://github.com/timbray/quamina/commit/5fe237bbadd003cac276cd0adc160d52263aeef6

```
  Incremental epsilon closure — results

  Plan executed (commits on embed-smalltable-research): spec 5fe237b → plan
  534327a → immutability gate 18547f6 → order-independence guard a6c44a9 →
  prune 5976a4f → review nits f96e39a (plus 20c980d research -cpuprofile
  flag).

  The win (10k-pattern research harness):

  ┌──────────────────────┬──────────────┬─────────────────────────────────┐
  │                      │ before prune │           after prune           │
  ├──────────────────────┼──────────────┼─────────────────────────────────┤
  │ patterns/sec         │ 133–140      │ 418 (3×)                        │
  ├──────────────────────┼──────────────┼─────────────────────────────────┤
  │ wall time            │ ~72 s        │ 24.6 s                          │
  ├──────────────────────┼──────────────┼─────────────────────────────────┤
  │ build-scale @ N=1000 │ 473 ms       │ 225 ms (~2.1×)                  │
  ├──────────────────────┼──────────────┼─────────────────────────────────┤
  │ closureForNfa flat   │ 8.46 s       │ 0.06 s (walk re-traversal gone) │
  └──────────────────────┴──────────────┴─────────────────────────────────┘
```

sayrer:
So, this is already getting close on AddPattern, and the footprint improvements improve the Matcher.

timbray:
Is this PR up-to-date with your latest?

sayrer:
No, this is on https://github.com/sayrer/quamina/tree/embed-smalltable-research, which has everything in this PR, your research PR, and further patches to see if the AddPattern cost is really necessary. I didn't want to pile them all in to one review, but I did want to see whether it's possible to avoid any unpleasant trade-offs, and that's looking promising.

timbray:
Oh, ok, I'll pull that and have a look around tomorrow sometime.

timbray:
OK, I got around to looking.  Looks good!  (Not looked at the code, just at the benchmarks).  A couple of mild concerns, not blockers but maybe you know what's going on:

1. In the research-main test, the number of states stays the same but the average closure fanout changes. ?!?
2. I notice in the graph that the shape of the AddPattern latency graph is really weird. Any idea what's going on?  I mean, the numbers are good, but it's still weird.

Picture1

Anyhow, what's the path forward - is this PR on top of my research/research-main PR or does it replicate mine?

sayrer:
> Anyhow, what's the path forward - is this PR on top of my research/research-main PR or does it replicate mine?

I cherry-picked it, so it has a different commit hash but preserves your authorship, I think you should ahead and merge your research PR, and then when I make PR of my branch it will merge cleanly.

I don't know the answers to the other two questions, but I'll have a look.

sayrer:
I got it all merged in. I left the docs/ folder in, in case you had design questions, but it doesn't have to stay.

I haven't gotten to your two questions yet.

Baseline 1ca29de (May 31)

```
  ┌───────────────────────────────────────---──┬──────────┬─────────┬───-─────┐
  │                Benchmark                   │ Baseline │ Current │    Δ    │
  ├──────────────────────────────────────---───┼──────────┼─────────┼──-──────┤
  │ CityLots                                   │   2.855µ │  2.868µ │   +0.5% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼-────────┤
  │ JsonFlattener_ContextFields                │   101.3n │  98.79n │   −2.5% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼─-───────┤
  │ JsonFlattener_MiddleNestedField            │   5.847µ │  5.854µ │   +0.1% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼──-──────┤
  │ JsonFlattener_LastField                    │   6.141µ │  6.094µ │   −0.8% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼-────────┤
  │ JsonFlattner_Evaluate_ContextFields        │   264.8n │  264.1n │   −0.3% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼─-───────┤
  │ JsonFlattner_Evaluate_MiddleNestedField    │   5.931µ │  6.245µ │   +5.3% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼──-──────┤
  │ JsonFlattner_Evaluate_LastField            │   6.199µ │  6.186µ │   −0.2% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼-────────┤
  │ NumberMatching                             │   217.0n │  213.0n │   −1.8% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼─-───────┤
  │ ShellstyleSimpleWildcard/a*b               │   208.9n │  185.8n │  −11.1% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼-────────┤
  │ ShellstyleSimpleWildcard/foo*bar           │   235.5n │  205.7n │  −12.7% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼─-───────┤
  │ ShellstyleSimpleWildcard/xyz               │   262.6n │  210.3n │  −19.9% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼-────────┤
  │ ShellstyleSimpleWildcard/he*lo             │   225.5n │  201.8n │  −10.5% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼─-───────┤
  │ ShellstyleNarrowInput/ascii/digits/8       │   635.8n │  521.9n │  −17.9% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼──-──────┤
  │ ShellstyleNarrowInput/ascii/digits/32      │   1.816µ │  1.352µ │  −25.5% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼-────────┤
  │ ShellstyleNarrowInput/ascii/digits/128     │   7.456µ │  5.159µ │  −30.8% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼-────────┤
  │ ShellstyleNarrowInput/ascii/lower/8        │   720.1n │  663.1n │   −7.9% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼─-───────┤
  │ ShellstyleNarrowInput/ascii/lower/32       │   2.076µ │  1.882µ │   −9.3% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼─-───────┤
  │ ShellstyleNarrowInput/ascii/lower/128      │   7.703µ │  6.297µ │  −18.3% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼-────────┤
  │ ShellstyleNarrowInput/ascii/cjk/8          │   1.902µ │  1.497µ │  −21.3% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼─-───────┤
  │ ShellstyleNarrowInput/ascii/cjk/32         │   5.571µ │  5.078µ │   −8.8% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼──-──────┤
  │ ShellstyleNarrowInput/ascii/cjk/128        │   20.66µ │  16.77µ │  −18.8% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼───-─────┤
  │ ShellstyleNarrowInput/cjk/digits/8         │   829.0n │  744.4n │  −10.2% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼────-────┤
  │ ShellstyleNarrowInput/cjk/digits/32        │   2.664µ │  2.132µ │  −20.0% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼─────-───┤
  │ ShellstyleNarrowInput/cjk/digits/128       │   11.68µ │  7.570µ │  −35.2% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼──────-──┤
  │ ShellstyleNarrowInput/cjk/lower/8          │   837.7n │  743.3n │  −11.3% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼─-───────┤
  │ ShellstyleNarrowInput/cjk/lower/32         │   2.702µ │  2.036µ │  −24.6% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼──-──────┤
  │ ShellstyleNarrowInput/cjk/lower/128        │   11.61µ │  7.329µ │  −36.9% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼───-─────┤
  │ ShellstyleNarrowInput/cjk/cjk/8            │   2.338µ │  1.888µ │  −19.2% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼────-────┤
  │ ShellstyleNarrowInput/cjk/cjk/32           │   8.385µ │  6.194µ │  −26.1% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼─────-───┤
  │ ShellstyleNarrowInput/cjk/cjk/128          │   37.35µ │  23.36µ │  −37.5% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼──────-──┤
  │ ShellstyleNarrowInput/mixed/digits/8       │   754.3n │  684.4n │   −9.3% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼───────-─┤
  │ ShellstyleNarrowInput/mixed/digits/32      │   2.514µ │  1.965µ │  −21.8% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼──────-──┤
  │ ShellstyleNarrowInput/mixed/digits/128     │   10.94µ │  6.719µ │  −38.6% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼────-────┤
  │ ShellstyleNarrowInput/mixed/lower/8        │   830.7n │  779.4n │   −6.2% │
  ├────────────────────────────────────────---─┼---───────┼─────────┼─────-───┤
  │ ShellstyleNarrowInput/mixed/lower/32       │   3.219µ │  2.208µ │  −31.4% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼──────-──┤
  │ ShellstyleNarrowInput/mixed/lower/128      │   16.02µ │  8.503µ │  −46.9% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼───-─────┤
  │ ShellstyleNarrowInput/mixed/cjk/8          │   2.495µ │  1.927µ │  −22.8% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼────-────┤
  │ ShellstyleNarrowInput/mixed/cjk/32         │   9.961µ │  6.479µ │  −35.0% │
  ├─────────────────────────────────────────---┼──────────┼─────────┼─────-───┤
  │ ShellstyleNarrowInput/mixed/cjk/128        │   47.49µ │  24.13µ │  −49.2% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼──────-──┤
  │ ShellstyleWidePatternsScaling/8            │   594.6n │  469.9n │  −21.0% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼───────-─┤
  │ ShellstyleWidePatternsScaling/16           │   979.3n │  673.5n │  −31.2% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼────────-┤
  │ ShellstyleWidePatternsScaling/32           │   1.672µ │  1.145µ │  −31.5% │
  ├──────────────────────────────────────---───┼──────────┼─────────┼───────-─┤
  │ ShellstyleWidePatternsScaling/64           │   3.410µ │  2.160µ │  −36.7% │
  ├───────────────────────────────────────---──┼──────────┼─────────┼───────-─┤
  │ ShellstyleWidePatternsScaling/128          │   7.403µ │  4.307µ │  −41.8% │
  ├────────────────────────────────────────---─┼──────────┼─────────┼─-───────┤
  │ ShellstyleWidePatternsScaling/256          │   16.77µ │  9.288µ │  −44.6% │
  ├────────────────────────────────────────────┼──────────┼─────────┼───-─────┤
  │ ShellstyleWidePatternsScaling/512          │   41.38µ │  21.70µ │  −47.6% │
  ├────────────────────────────────────────────┼──────────┼─────────┼───-─────┤
  │ ShellstyleSimpleWildcardScaling/1          │   245.7n │  194.5n │  −20.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ ShellstyleSimpleWildcardScaling/4          │   253.9n │  197.9n │  −22.1% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ ShellstyleSimpleWildcardScaling/8          │   256.3n │  197.9n │  −22.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼───-─────┤
  │ ShellstyleSimpleWildcardScaling/16         │   256.4n │  201.4n │  −21.5% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ ShellstyleSimpleWildcardScaling/26         │   255.5n │  200.8n │  −21.4% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ ShellstyleZWJEmoji/4                       │   3.069µ │  2.273µ │  −25.9% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ ShellstyleZWJEmoji/8                       │   4.546µ │  3.192µ │  −29.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─-───────┤
  │ ShellstyleZWJEmoji/16                      │   6.237µ │  4.443µ │  −28.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ ShellstyleZWJEmoji/32                      │   9.828µ │  7.216µ │  −26.6% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ ShellstyleZWJEmoji/64                      │   19.96µ │  13.29µ │  −33.4% │
  ├────────────────────────────────────────────┼──────────┼─────────┼───-─────┤
  │ ShellstyleMultiMatch                       │   18.57µ │  14.52µ │  −21.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ Nfa2Dfa/single_star                        │   69.77µ │  34.52µ │  −50.5% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─-───────┤
  │ Nfa2Dfa/two_stars                          │   169.9µ │  76.75µ │  −54.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ Nfa2Dfa/three_stars                        │   40.93µ │  20.77µ │  −49.3% │
  ├────────────────────────────────────────────┼──────────┼─────────┼──-──────┤
  │ Nfa2Dfa/five_stars                         │   59.09µ │  28.92µ │  −51.1% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─-───────┤
  │ Nfa2Dfa/eight_stars                        │   83.38µ │  39.78µ │  −52.3% │
  ├────────────────────────────────────────────┼──────────┼─────────┼───-─────┤
  │ 8259Example                                │   4.025µ │  3.512µ │  −12.7% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─-───────┤
  │ ShellStyleBuildTime                        │   20.45m │  14.33m │  −29.9% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─-───────┤
  │ TablePointerDedup/6-regexps-12-shell       │   4.414µ │  3.604µ │  −18.4% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ TablePointerDedup/20-nested-regexps        │   499.0n │  447.2n │  −10.4% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ TablePointerDedup/deeply-nested            │   538.3n │  467.5n │  −13.2% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ TablePointerDedup/overlapping-char-classes │   393.3n │  344.7n │  −12.4% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ TablePointerDedup/shell+deep-overlap       │   6.043µ │  4.848µ │  −19.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ExactString                       │   108.9n │  97.28n │  −10.7% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_SingleShellstyle                  │   256.2n │  202.7n │  −20.9% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ManyOverlappingWildcards/N=8      │   705.7n │  583.4n │  −17.3% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ManyOverlappingWildcards/N=16     │   1.473µ │  1.166µ │  −20.8% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ManyOverlappingWildcards/N=32     │   2.991µ │  2.257µ │  −24.5% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ManyOverlappingWildcards/N=64     │   6.733µ │  4.510µ │  −33.0% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ManyOverlappingWildcards/N=128    │   14.34µ │  9.935µ │  −30.7% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_RegexAlternation                  │   36.67n │  36.19n │   −1.3% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_LiteralInRegex                    │   36.61n │  36.97n │   +1.0% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_QuantifiedCharClass               │   37.11n │  36.97n │   −0.4% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ManyAnchoredRegex                 │   38.74n │  36.38n │   −6.1% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_DeepEpsilonNest                   │   37.83n │  36.16n │   −4.4% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_CacheThrashing                    │   205.3n │  166.1n │  −19.1% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ParallelMatchers/G=8              │   2.088µ │  1.532µ │  −26.6% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ParallelMatchers/G=16             │   2.057µ │  1.573µ │  −23.5% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ParallelMatchers/G=32             │   2.122µ │  1.488µ │  −29.9% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ Workload_ParallelMatchers/G=64             │   2.122µ │  1.813µ │  −14.6% │
  ├────────────────────────────────────────────┼──────────┼─────────┼─────────┤
  │ geomean                                    │   2.277µ │  1.747µ │ −23.31% │
  └────────────────────────────────────────────┴──────────┴─────────┴─────────┘

Memory
  
  B/op and allocs/op are essentially unchanged across the board
  (geomean −0.49% B/op, −0.75% allocs). The only movement is Nfa2Dfa/*, which
  allocates slightly less (e.g. two_stars 222→202 allocs, 98.1→97.6 KiB).
  Everything else is identical.

Takeaways
  - embed-smalltable is a broad win — geomean −23% on time, with no memory 
    regression.
  - Biggest gains are in DFA construction (Nfa2Dfa ~−50–55%) and wide/scaling
    shellstyle matching (up to −49% at high pattern counts), consistent with
    the embedded smalltable reducing pointer indirection during traversal.
  - Flater paths unchanged — JSON flattening, exact-string, and the tiny regex
    workloads (~36ns) are flat, as expected since they don't stress smalltable
    traversal. One blip: Evaluate_MiddleNestedField at +5.3%, but with n=3 
    and high variance benchstat marks it non-significant (~).

⚠️  Statistical caveat: every row is flagged ~ (p≥0.05) because -count=3 is 
below benchstat's 6-sample threshold for confidence intervals. The deltas
are large and directionally consistent, but if you want publishable numbers
I can re-run with -count=10 (≈40 min for the new branch alone).
```

sayrer:
> In the research-main test, the number of states stays the same but the average closure fanout changes. ?!?

The fanout numbers are because this branch uses a sentinel value to avoid work. We could do this in `memory_cost.go`:

```
  le := int64(len(state.epsilonClosure))
  if le == 0 {
      le = 1 // self-only sentinel: the closure is {self}, cardinality 1
  }
  stats.fanouts += le
```

sayrer:
The jitttery chart is because of the ` O(N²)→O(N)` win in 5976a4f.  I had Cluade bisect it. This change reduced the uniform work in each window, leaving only the variations in the input.  The first set of pictures in this PR both show smooth slopes, but this is because they were doing too much constant work. So, I might suggest changing the way m/AddP is done.

n.b. these are different data sets, but I had them around

Today's chart:

Screenshot 2026-06-05 at 5 17 41 PM

A diffferent way:

Screenshot 2026-06-05 at 5 18 42 PM

probably for a new PR

timbray:
OK, this PR now stable from your POV?

sayrer:
yes

dedup_key.go
```diff
@@ -0,0 +1,26 @@
...
+type tableShareKey struct {
+	stepsData unsafe.Pointer
+	stepsLen  int
+}
+
+func newTableShareKey(t *smallTable) tableShareKey {
+	return tableShareKey{
+		stepsData: unsafe.Pointer(unsafe.SliceData(t.steps)),
+		stepsLen:  len(t.steps),
+	}
```
timbray, line 30: (interim note to self): Not obvious why you need the length - if the two pointers are the same, shouldn't that be enough?

dedup_key_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestTableShareKey_SharedBackings(t *testing.T) {
+	// Construct one smallTable, value-copy it (simulating post-embed share).
+	src := smallTable{
+		ceilings: []byte{'a', 'b', byte(byteCeiling)},
+		steps:    []*faState{nil, nil, nil},
+	}
+	copy1 := src
+	copy2 := src
+	if newTableShareKey(&copy1) != newTableShareKey(&copy2) {
```
timbray, line 15: Haha, I didn't actually know you could test struct equality like this.

epsi_closure.go
```diff
@@ -1,70 +1,107 @@
...
-// permanent presence on every smallTable was wasted steady-state memory.
-// They now live in a per-call side table that is discarded when
-// epsilonClosure returns.
+import "sync"
+
+// selfOnlyClosure is the shared sentinel for a closure equal to {self}: non-nil
+// (distinct from nil = "not computed") and zero-length (so consumers take their
+// len==0 self-only branch). The empty composite literal points at runtime
+// zerobase, so this allocates nothing.
+var selfOnlyClosure = []*faState{}
```
timbray, line 60: Wondering about the trade-off here. In the particular case where len(closure) == 1, you save 32 bytes I guess - 3 vals for slice + one pointer at the cost of slightly more complex code. Or am I missing something?

epsi_closure.go
```diff
@@ -1,70 +1,107 @@
...
+// pooled side table (closureBuffers.tables).
+//
+// tableMark is stored by value so marking a share group costs no per-entry
+// heap allocation.
 type tableMark struct {
-	lastVisitedGen uint32
-	closureGen     uint32
-	closureRep     *faState
+	closureGen uint64
+	closureRep *faState
```
timbray, line 80: The name "closureRep" is not that self-explanatory and there used to be a comment explaining what it was, a bit sad to have lost that.

epsi_closure.go
```diff
@@ -1,70 +1,107 @@
...
 }
 
-// closureBuffers carries per-epsilonClosure-call scratch. The two maps
-// replace build-time fields that used to sit on smallTable/faState;
-// they live only for the duration of the closure computation.
+// closureBuffers carries the scratch for epsilon closure computation. It is
+// pooled (see closureBufferPool) and reused across epsilonClosure calls, so
```
timbray: Since this is of necessity single-threaded - called from AddPattern, which is mutexed - does sync.Pool add any value?  It's not that cheap, in my experience.

epsi_closure.go
```diff
@@ -1,70 +1,107 @@
...
+// tracking is generation-based: gen only ever increases, so stale map
+// entries from a previous use are simply older than the current generation
+// and need no clearing.
 type closureBuffers struct {
-	gen           uint32                     // bumped by closureForNfa (NFA walk dedup) and the closureForState post-pass (table-pointer dedup)
-	closureSetGen uint32                     // snapshot of gen used by traverseEpsilons to dedup faState visits within one closure
-	closureList   []*faState                 // reusable accumulator for the state list before the dedup post-pass
-	tables        map[*smallTable]*tableMark // per-call side-table for smallTable scratch (lastVisitedGen, closureRep)
-	states        map[*faState]uint32        // per-faState last-visited generation, used by traverseEpsilons
+	gen           uint64                      // monotonic counter; bumped by closureForState's two dedup phases
```
timbray, line 91: This is awfully complex. Maybe a big comment somewhere explaining at a high level how epsilon closure building works? Could probably copy that in from your Claude gudance in plans/?

epsi_closure.go
```diff
@@ -1,70 +1,107 @@
...
-func (b *closureBuffers) tableMarkOf(t *smallTable) *tableMark {
-	m, ok := b.tables[t]
-	if !ok {
-		m = &tableMark{}
-		b.tables[t] = m
-	}
-	return m
+// closureBufferPool reuses closureBuffers (and their maps) across the many
+// epsilonClosure calls a build performs, eliminating per-call map allocation.
+// The pool is concurrency-safe, and sync.Pool drops its contents on GC, so
```
timbray: Not sure concurrency is an issue in anything that descends from AddPattern()

memory_cost.go
```diff
@@ -49,7 +49,9 @@ func cmStateStats(state *faState, stats *matcherStats, pp printer) {
 		}
 	}
 	for _, eps := range state.table.epsilons {
-		cmStateStats(eps, stats, pp)
+		if eps != nil {
```
timbray: Red flag - how can there be a nil entry in smallTable.epsilons?

nfa.go
```diff
@@ -262,12 +259,12 @@ func traverseDFA(table *smallTable, val []byte, transitions []*fieldMatcher) []*
 // collected in the nextStates list.  The bufs structure contains three buffers, one each for
 // currentStates, nextStates, and the epsilon closure of one particular state. These are re-used
 // and should grow with use and minimize the need for memory allocation.
-func traverseNFA(table *smallTable, val []byte, transitions []*fieldMatcher, bufs *nfaBuffers) []*fieldMatcher {
+func traverseNFA(start *faState, val []byte, transitions []*fieldMatcher, bufs *nfaBuffers) []*fieldMatcher {
 	currentStates := bufs.getBuf1()
 	// The start state always has a trivial epsilon closure (just itself) because
 	// all Quamina automata begin by matching the opening quote (0x22). The start
```
timbray: Oh wait, that's not true.  The values true, false, and null don't have quotes, nor do numbers.  I think you still have a trivial closure though?

nfa.go
```diff
@@ -285,6 +282,16 @@ func traverseNFA(table *smallTable, val []byte, transitions []*fieldMatcher, buf
 			utf8Byte = valueTerminator
 		}
 		for _, state := range currentStates {
+			if len(state.epsilonClosure) == 0 {
+				// self-only closure: process the state itself
```
timbray, line 295: All these little code fragments to work around special-casing the len()==1 sentinel. A bit dubious about this trade-off. Wait, I already said that above.

nfa.go
```diff
@@ -376,9 +389,9 @@ func simplifyCollect(s *faState, visited map[*faState]bool, targets []*faState)
 // minimal or even avoids being wasteful.
 // INVARIANT: neither argument is nil
 // INVARIANT: To be thread-safe, no existing table can be updated except when we're building it
-func mergeFAs(table1, table2 *smallTable, pp printer) *smallTable {
-	state1 := &faState{table: table1}
-	state2 := &faState{table: table2}
+func mergeFAs(table1, table2 *smallTable, pp printer) smallTable {
+	state1 := &faState{table: *table1}
+	state2 := &faState{table: *table2}
```
timbray, line 403: Since you've done all the work of changing all this stuff to start at an faState not a smallTable (yay) wondering if you could just pass states to mregeFAs.

nfa.go
```diff
@@ -420,22 +433,22 @@ func mergeFAStates(state1, state2 *faState, keyMemo map[faStepKey]*faState, pp p
...
 		combined.table.epsilons = simplifySplices(state1, state2)
 		keyMemo[mKey] = combined
 		return combined
 	}
 
 	combined.fieldTransitions = append(state1.fieldTransitions, state2.fieldTransitions...)
 
-	pp.labelTable(combined.table, fmt.Sprintf("%d∎%d",
-		pp.tableSerial(state1.table), pp.tableSerial(state2.table)))
+	pp.labelTable(&combined.table, fmt.Sprintf("%d∎%d",
```
timbray, line 463: In some future golden era where I never need to debug this stuff again, we can lose the prettyprinter here, but maybe that time is not yet.

state_lists.go
```diff
@@ -38,23 +36,16 @@ func newStateLists() *stateLists {
 // which either has already been computed for the set or is created and empty, and
 // a boolean indicating whether the DFA state has already been computed or not.
 func (sl *stateLists) intern(list []*faState) ([]*faState, *faState, bool) {
```
timbray, line 38: I'll be interested in what the profiler says.  This used to dominate nfa2Dfa performance…

value_matcher.go
```diff
@@ -23,7 +23,7 @@ type valueMatcher struct {
 	updateable atomic.Pointer[vmFields]
 }
 type vmFields struct {
-	startTable          *smallTable
+	startState          *faState
```
timbray: My favorite change in this PR. The field name could probably be sorted just to "start" if you feel like it.

value_matcher.go
```diff
@@ -114,38 +114,46 @@ func (m *valueMatcher) addTransition(val typedVal, printer printer) *fieldMatche
 	// no dodges, we have to build an automaton to match this value
 	var nextField *fieldMatcher
 
-	var newFA *smallTable
+	// newFA holds the newly-built automaton. Most builders return a smallTable
+	// value to be wrapped in an faState; makeRegexpNFA and a few NFA builders
```
timbray, line 118: Might want to go back and make all the FA generators return faState not smallTable.  Can wait.

value_matcher.go
```diff
@@ -114,38 +114,46 @@ func (m *valueMatcher) addTransition(val typedVal, printer printer) *fieldMatche
...
+		printer.labelTable(&newFAState.table, "RX start")
 	default:
 		panic("unknown value type")
 	}
 
 	// there's already a table, thus an out-degree > 1
-	if fields.startTable != nil {
-		fields.startTable = mergeFAs(fields.startTable, newFA, printer)
+	if fields.startState != nil {
+		mergedTable := mergeFAs(&fields.startState.table, &newFAState.table, printer)
```
timbray: (maybe suggested this earlier…) possible to turn the args into faState rather than smallTable?

epsi_closure.go
```diff
@@ -96,26 +133,52 @@ func closureForState(state *faState, bufs *closureBuffers) {
...
 				continue
 			}
 		} else {
 			mark.closureGen = dedupGen
 			mark.closureRep = s
+			bufs.tables[key] = mark
 		}
 		closure = append(closure, s)
 	}
+	if !state.table.isEpsilonOnly() && len(closure) == 1 {
```
timbray, line 259: I note in the test coverage that this if branch is never taken. Is that worrying?

timbray (commented):
There are three distinct things in here:
1. Mostly have things start on faState rather than smallTable, including the field in valueMatcher.  I have wanted to do this for like a year, so this is great. I think this job perhaps isn't 100% done (see comments on arguments to mergeFAs and valueMatcher generally) but probably don't need to block on that.  I may be able to go back and drastically simplify {"exists":true} matching depending simply on whether vm.start is nil or not.
2. Turn state.table from *smallTable to smallTable.  Excellent.
3. Refactor epsilonClosure.  Frankly I don't understand what's happening here and there are a lot of fields to support it. I've suggested adding a large overview comment, probably sourced from your Claude dialog. I'd prefer not to add the Claude stuff in docs/ to the project as a thing in itself.

Anyhow, this is huge and almost ready, thanks!

epsi_closure.go
sayrer: Added a comment for this one.

epsi_closure.go
sayrer: I had Claude go back and read why we did this:

"  The byte math is off. The 24-byte slice header (ptr/len/cap) lives inline in the faState struct — the epsilonClosure field is a header either
  way, so you can't save it. What the sentinel actually saves is the heap-allocated backing array: for []*faState{self} that's 1 pointer → rounds
  to a size class, plus per-object GC-scan and allocator metadata. So it's ~8–16 bytes plus an allocation, not 32 bytes.
  
  It's not "the particular case where len==1" — that's the common case. From the measured closure-size distribution (shellstyle, N=2000) in the
  design doc:

```
  ┌────────┬────────────┬─────┬─────┬─────┬───────┐
  │ states │ total mass │ avg │ p50 │ p90 │  max  │
  ├────────┼────────────┼─────┼─────┼─────┼───────┤
  │ 14,573 │ 98,419     │ 6.8 │ 1   │ 2   │ 1,236 │
  └────────┴────────────┴─────┴─────┴─────┴───────┘
```

  p50 = 1 means the majority of states are self-only. So this eliminates an allocation for most states in the automaton, while a handful of hub
  states (max 1,236) carry nearly all the real mass and are untouched.

  It's permanent memory, not transient. epsilonClosure is consulted at match time, so each slice lives for the matcher's whole lifetime. So you're
  saving N allocations' worth of resident, GC-scanned memory permanently — not just build-time churn.

  The measured payoff is match-time speed, not bytes. From the benchmark commit (benchstat n=6): shellstyle match −12% sec/op geomean, up to −34%
  on wide patterns; research harness +34–37% matches/sec at 5k–10k patterns; −1.8% model bytes; build neutral. Taking the len==0 branch for the
  common state avoids the backing-array pointer-chase on the hot match path — that's the real win.
  
  Why not the simpler "just drop self everywhere": the doc records that a prior spike doing exactly that hung TestToxicStack/TestRegexpValidity —
  pulling self out of a multi-member closure defeats the table-pointer dedup (self shares a smallTable with closure members) → exponential blowup
  on adversarial regexps. So self must stay inside multi-member closures; only the self-only case (nothing to dedup against) is safe to specialize.
  That asymmetry is the reason for the "slightly more complex code."

  The current code comment only explains the mechanism (zerobase, zero-alloc). I could add one line pointing at the design doc and noting the
  "common case / match-time win" rationale, which would preempt exactly this question for the next reader."

epsi_closure.go
sayrer: https://github.com/timbray/quamina/pull/535/changes/83492f84e894dd6583a11dfd2afe3ea6fad7daa7

value_matcher.go
sayrer: Yes, let's defer this one. Start states can be spinners, so it belongs in its own change rather than the embed cleanup.

state_lists.go
sayrer: Profiled BenchmarkNfa2Dfa (M5, CPU profile). Intern still dominates nfa2Dfa, ~65% of its CPU (1.94s of 3.0s cum).

But the cost has moved. After this branch's changes (sort+compact dedup, reusable scratch buffers, PutUint64 key-build, alloc-free string(keyBuf) lookup), the marshalling is now cheap and the hot line is the map lookup itself:
  
  ```
┌───────────────────────────────┬───────────────────────┬─────────────────────────┐
  │             line              │         cost          │          what           │
  ├───────────────────────────────┼───────────────────────┼─────────────────────────┤
  │ sl.entries[string(sl.keyBuf)] │ 1.08s — 56% of intern │ string-hash + map probe │
  ├───────────────────────────────┼───────────────────────┼─────────────────────────┤
  │ slices.SortFunc               │ 280ms                 │ canonical-key sort      │
  ├───────────────────────────────┼───────────────────────┼─────────────────────────┤
  │ append(sortBuf[:0], list...)  │ 250ms                 │ copy input to scratch   │
  ├───────────────────────────────┼───────────────────────┼─────────────────────────┤
  │ key-build loop (PutUint64)    │ 130ms                 │ per-state               │
  ├───────────────────────────────┼───────────────────────┼─────────────────────────┤
  │ rest (compact / miss alloc)   │ ~140ms                │                         │
  └───────────────────────────────┴───────────────────────┴─────────────────────────┘
```
 
So the code around the lookup is about as tight as it gets; the residual is the irreducible set-dedup hash. Further wins would need a cheaper key/hash or fewer intern calls, not trimming the surrounding code. (The big madvise/GC in the raw profile is n2dNode building DFA states across nfa2Dfa, not intern.)

File: nfa.go (hunk 10 above)
Comment (sayrer) [reply nfa.go:295]: Unfortunately, this is load bearing. Tried a "cleaner" version and got a big regression.

```
  ┌─────────────────────────────────────────────┬───────────────┐
  │                  benchmark                  │       Δ       │
  ├─────────────────────────────────────────────┼───────────────┤
  │ ShellstyleWidePatternsScaling (8…512)       │ +22% to +27%  │
  ├─────────────────────────────────────────────┼───────────────┤
  │ ShellstyleMultiMatch                        │ +26%          │
  ├─────────────────────────────────────────────┼───────────────┤
  │ Workload_ManyOverlappingWildcards (N=8…128) │ +7% to +13%   │
  ├─────────────────────────────────────────────┼───────────────┤
  │ Workload_SingleShellstyle                   │ +8%           │
  ├─────────────────────────────────────────────┼───────────────┤
  │ RegexAlternation / DeepEpsilonNest          │ ~neutral      │
  ├─────────────────────────────────────────────┼───────────────┤
  │ geomean                                     │ +15.7% slower │
  └─────────────────────────────────────────────┴───────────────┘
  
```

File: value_matcher.go (hunk 16 above)
Comment (sayrer) [reply value_matcher.go]: It's doable but behavior-affecting via root spinners, so it warrants its own benchmarked change rather than riding along here.

File: nfa.go (hunk 11 above)
Comment (sayrer) [reply nfa.go:403]: It's doable but behavior-affecting via root spinners, so it warrants its own benchmarked change rather than riding along here.

**Comment by sayrer** [comment]
OK, I think I got everything or explained them at least.

**Comment by timbray** [comment]
I'll go through this a little later, but a quick note on first glance: If you go through quamina.AddPattern() you can't have a spinner on the start state, because regexps can only apply to strings, so the first step in the regexp is always a literal `"` or 0x22.   In testing, it's easy to build a wildcard or regexp that starts with `*` or `.*`, but not in actual use.

Arguably Quamina always retaining the enclosing quotations adds 2 extra steps to FA traversal, but it simplifies a  bunch of other things.

File: epsi_closure.go (hunk 3 above)
Comment (timbray) [reply epsi_closure.go:60]: OK

File: docs/superpowers/plans/2026-05-31-incremental-epsilon-closure.md [hunk 18]
```diff
@@ -0,0 +1,338 @@
+# Incremental Epsilon Closure (Walk Pruning) Implementation Plan
```
timbray: Please remove the Claude stuff.

epsi_closure.go
```diff
@@ -1,70 +1,168 @@
...
-		m = &tableMark{}
-		b.tables[t] = m
-	}
-	return m
+// closureBufferPool reuses closureBuffers (and their maps) across the many
+// epsilonClosure calls a build performs, eliminating per-call map allocation.
+// A single build (everything under one AddPattern) is single-threaded, so the
+// buffers are never shared concurrently within a build. sync.Pool is used
+// rather than a plain free list for two reasons: this is a package-level
+// global, so independent matcher builds running in separate goroutines draw
```
timbray: You mean "independent matcher builds" on an entirely separate Quamina instance?  In the same Go program? That seems wildly unlikely. As the comment makes clear, you can't have concurrent matcher builds on the same instance.

I still can't see why it wouldn't be simpler & probably faster just to have a single set of closureBuffers around per instance and have a resetClosureBuffers() call to clear the maps at the start of AddPattern. There's no reason ever to GC this stuff unless you know for sure you're never going to call AddPattern again.

timbray (commented):
Only remaining quibble is the sync.Pool() stuff.   Also see earlier note about root spinners (can't happen)

timbray:
Now reading https://pkg.go.dev/sync#Pool and trying to figure out what I think.

sayrer:
> I'll go through this a little later, but a quick note on first glance: If you go through quamina.AddPattern() you can't have a spinner on the start state, because regexps can only apply to strings, so the first step in the regexp is always a literal `"` or 0x22. In testing, it's easy to build a wildcard or regexp that starts with `*` or `.*`, but not in actual use.
> 
> Arguably Quamina always retaining the enclosing quotations adds 2 extra steps to FA traversal, but it simplifies a bunch of other things.

I wrote a buggy test that starts with `*`. will fix.

timbray:
The more I think about it, the less comfortable I get with sync.Pool.  Apparently it randomly ejects things from cache in GC operations. But our pattern is super-deterministic. For each AddPattern, we

1. Clear the maps in the cache
2. add lots of stuff and consult it frequently while we build automata and epsilon closures
3. Are done with it (could clear it here)

I'm not 100% sure what the effect is of syncPool() deciding to expire stuff while we're at stage 2 in the list above, but it's not going to be helpful. 

Could we take out sync.Pool() and just have a simple use/clear rhythm?

sayrer:
> Could we take out sync.Pool() and just have a simple use/clear rhythm?

```
  - Full suite green; -race clean on the Copy()/concurrent-AddPattern paths 
    (the key risk of dropping the concurrency-safe pool).

  - Build +22% vs the pool (732 vs 599 patterns/sec, clean repeated harness runs) — and
    notably the naïve never-clear version was −7%, so the per-AddPattern clear() is what
    makes it a win (keeps the scratch maps holding one build's working set instead of
    growing to O(all states).

  - Match path untouched; match benchmarks unchanged.
  
  - Generation counters retained (they avoid clearing within a build;
    reset() bounds size across builds).
```

timbray:
This is about the fifth time I've observed sync.Pool being more expensive than seems reasonable. 

Heh, Claude needs to really internalize the Quamina concurrency model.  "-race clean" is not a surprise. 

Anyhow, this looks good, I'll do one more sweep but I think we can merge.  

One question: GitHub is whining about an unsigned commit.  Is this an admin error you remember? if so, no problem. Otherwise, maybe an attacker?

value_matcher.go
```diff
@@ -114,38 +114,45 @@ func (m *valueMatcher) addTransition(val typedVal, printer printer) *fieldMatche
 	// no dodges, we have to build an automaton to match this value
 	var nextField *fieldMatcher
 
-	var newFA *smallTable
+	// newFA holds the newly-built automaton. Most builders return a smallTable
+	// value to be wrapped in an faState; makeRegexpNFA and a few NFA builders
+	// return *faState directly. After this switch, newFAState is the start
+	// faState for the new automaton.
+	var newFAState *faState
```
timbray, line 121: At some future point I'm going to rename this back to newFA but first I'm going to merge this.

timbray (approved):
This has become da big one, happy to land it. 

At some future point it may be worth trying to simplify epsi_closure.go which is now maybe the most complicated piece of logic in the whole project. But it'll do for now.

sayrer:
> One question: GitHub is whining about an unsigned commit. Is this an admin error you remember? if so, no problem. Otherwise, maybe an attacker?

It's a new laptop (no Migration Assistant) and I forgot to do the ssh key. It's fixed now, but once one commit slips in, it won't verify the following ones in a PR.

timbray:
Got it, thanks. Merging…

Issue #530, opened 2026-05-27
Stats for matcher with 5000 wildcards


timbray:
I have failed to implement a memory-budget mechanism that is (a) deterministic and (b) has acceptable performance.   So I investigated another approach where I implemented a `GetUsageStats()` API that returns the number of states, the number of bytes, the the average and maximum closure fanout, and the performance for `AddPattern` and `MatchesForEvent`.  The spreadsheet is [here](https://docs.google.com/spreadsheets/d/1qEE89FBAtHFqkQum0bsD8_Be2UTE9BmVMmLVXmb_T4M/edit?usp=sharing). I extracted a few graphs. In each case the X axis is the number of patterns added, 0-5000. The patterns are those from `TestShellStyleBuildTime`.

<img width="223" height="128" alt="Image" src="https://github.com/user-attachments/assets/9edf655b-41ca-44e1-ab71-a8926929cb88" />

<img width="223" height="128" alt="Image" src="https://github.com/user-attachments/assets/310fd100-6614-44b2-a610-8087ec878474" />

<img width="218" height="128" alt="Image" src="https://github.com/user-attachments/assets/70255c5c-c553-48c9-85c7-153c350bb655" />

<img width="225" height="128" alt="Image" src="https://github.com/user-attachments/assets/ec132475-6ffa-4f1d-a636-5f774c1e312a" />

<img width="223" height="128" alt="Image" src="https://github.com/user-attachments/assets/871ef775-83b7-4023-9075-de9e0bce0af1" />

<img width="225" height="124" alt="Image" src="https://github.com/user-attachments/assets/2d3ed31f-0d08-40b3-bac0-4c942b375350" />

<img width="225" height="124" alt="Image" src="https://github.com/user-attachments/assets/de764cfc-e521-46c6-b91a-5017f37dd7fc" />

timbray:
So the runtime cost of both adding and matching is mildly superlinear and is reasonably well approximated by the matcher byte-count (which BTW is now exact and correct, following on laborious hand-checking).

timbray:
BTW I ran again with 10K patterns to see if the benchmark was missing any elbows in the curve, and apparently not.

timbray:
Haha, and now I put in DFA conversion and the benchmark has been running for 45 minutes...

timbray:
In fact, the AddPattern slowdown is ferocious. I eventually settled for a run with only 300 patterns, the graphs tell the story no kidding at all.

[...]

PR #541, opened 2026-06-11
api: draft build-mode API

timbray:
Open for comments, please pile on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,12 +201,32 @@ yet implemented.***
 The `SetMemoryBudget()` and `GetMemoryBudget()` APIs are deprecated. As implemented
 they were too expensive and occasionally nondeterministic.
 
+### Comfort vs Speed
+
+```go
+func (q *Quamina) SetMatcherBuildMode(mode MatcherBuildMode)
+func (q *Quamina) GetMatcherBuildMode() MatcherBuildMode
+```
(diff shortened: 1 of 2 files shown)


timbray:
Hey @jsmorph you still around?  Check the changes in this draft, also #530. I've got a much bigger PR cooking that will actually implement this, including the DFA conversion.  Turns out this is going to change the requirements for the LiveStorage type because unless I'm missing something it's gong to have to remember the BuildMode for each pattern and restore them in the same order they were applied. Which is enough of a semantic change that maybe we need a new OrderedLiveStorage type or suchlike.  Anyhow, would appreciate your thoughts on the matter before I dive in and start changing things.

timbray:
Further thoughts. I think we do need a LivePatterns2 class that stores either pairs of patterns/BuildMode or something logically equivalent then a Pruner2 which differs only in using a LivePatterns2. Then `SetMatcherBuildMode()` will check whether the current Matcher is either a coreMatcher or Pruner2 and refuse to proceed if otherwise.

timbray:
Hah, cancel the concern.  The `WithPatternStorage()` option is not implemented, and the docs say so.  So we can just change the implementation.

timbray:
Superseded by #543

PR #546, opened 2026-06-16
kaizen: fix typos in README "Compiling for specific architectures"

sayrer:
Missed the ones I was originally looking at. These are all my mistakes.


timbray (approved):

timbray:
I have a list of this kind of thing too, will drop a PR. Most hilarious is that in smallTable, step and dStep are now identical unless I'm missing something.

PR #547, opened 2026-06-17
kaizen: Closure visited gen

sayrer:
speeds up build time by 1.77x


sayrer:
The first commit is from a few weeks ago, before I had fixed the ssh key on my laptop, that's why this isn't verified.

timbray (commented):
Not going to take this one for a while.  The epsilon-closure code has grown a major code smell, very opaque, so much deduping.  I need to spend time taking a close look and walkthrough, I have this hunch that we're missing something.

sayrer:
All good. My hunch was that the code is complicated because the problem is difficult. But if it can be cleaner, that would be better.

PR #568, opened 2026-06-26
pat: treat {0,} quantifier as * to avoid nil-pointer panic

### LINKED ISSUE #558: regexp containing a{0,} can provoke a nil-pointer crash
Code is in regexp_nfa.go:131-144, leaves `lastState` at nil, later crash.  Need to fix and document whether `{0,}` is allowed and if so what it means.

SAY-5:
Fixes #558. The `isMinimumOnly` branch in `faFromQuantifiedAtom` builds a chain of mandatory steps and then loops the last one back, but for `{0,}` the loop runs zero times, so `lastState` stays nil and the append on it panics. `{0,}` has no mandatory steps and means the same thing as `*`, so I handle that case the same way the `*` branch does.

Added an `x{0,}y` entry to `regexpSamples` (matches `y`/`xy`/`xxxy`, rejects `x`/`z`/`xx`); without the fix it panics during NFA construction.

### DIFF
--- a/regexp_nfa.go
+++ b/regexp_nfa.go
@@ -124,6 +124,14 @@ func faFromQuantifiedAtom(branch regexpBranch, index int, finalStep *faState, pp
 		state.table.epsilons = []*faState{nextState}
 
 	case atom.isMinimumOnly():
+		// {0,} has no mandatory steps, so it means the same thing as *
+		if atom.quantMin == 0 {
+			state = &faState{}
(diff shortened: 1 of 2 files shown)


timbray (approved):
Actually I realized how short it is, lgtm, thanks.

In future please sign your PRs, not with commit -s but by setting up a keypair with GitHub, years since I did it so I can't remember how but pretty straightforward I think.

PR #573, opened 2026-08-11
fix: Accept supplementary Unicode in regexp character classes

### LINKED ISSUE #564: isCCchar upper bound typo
`0x10fff` should be `0x10ffff` astral-plane chars wrongly rejected. There are astral-plane tests but apparently they all fall below 10fff?

Boulea7:
Fixes #564.

Corrects the regexp character-class upper bound so supplementary Unicode code points are accepted. Adds boundary coverage and an external regression test for direct classes and Unicode properties.

Tested locally with go test -race -count 1 and golangci-lint run.

### DIFF
--- a/external_test.go
+++ b/external_test.go
@@ -1,7 +1,9 @@
 package quamina_test
 
 import (
+	"fmt"
 	"testing"
+	"unicode"
 
 	"quamina.net/go/quamina/v2"
 )
@@ -48,5 +50,47 @@ func TestDifferentFlattener(t *testing.T) {
 	}
 	if len(matches) != 1 || matches[0] != "xyz" {
 		t.Error("missed!")
+	}
+}
+
+func TestRegexpCharacterClassAboveU10FFF(t *testing.T) {
+	const mathematicalBoldDigitZero = '\U0001D7CE'
+	if !unicode.Is(unicode.Nd, mathematicalBoldDigitZero) {
+		t.Fatalf("U+%04X is not a decimal digit", mathematicalBoldDigitZero)
+	}
+
+	q, err := quamina.New()
+	if err != nil {
+		t.Fatalf("New: %v", err)
+	}
+	patterns := []struct {
+		name    string
+		pattern string
+	}{
+		{name: "literal character class", pattern: fmt.Sprintf(`{"value": [{"regexp": "[%c]"}]}`, mathematicalBoldDigitZero)},
+		{name: "decimal digit property", pattern: `{"value": [{"regexp": "~p{Nd}"}]}`},
+	}
+	wantMatches := make(map[string]bool, len(patterns))
+	for _, pattern := range patterns {
+		if err := q.AddPattern(pattern.name, pattern.pattern); err != nil {
+			t.Fatalf("AddPattern(%q): %v", pattern.name, err)
+		}
+		wantMatches[pattern.name] = true
+	}
+
+	event := fmt.Appendf(nil, `{"value": "%c"}`, mathematicalBoldDigitZero)
+	matches, err := q.MatchesForEvent(event)
+	if err != nil {
+		t.Fatalf("MatchesForEvent: %v", err)
+	}
+	for _, match := range matches {
+		name, ok := match.(string)
+		if !ok || !wantMatches[name] {
+			t.Fatalf("got unexpected match %v", match)
+		}
+		delete(wantMatches, name)
+	}
+	if len(wantMatches) != 0 {
+		t.Fatalf("missing regexp pattern matches: %v", wantMatches)
 	}
 }
--- a/external_test.go
+++ b/external_test.go
@@ -1,9 +1,7 @@
 package quamina_test
 
 import (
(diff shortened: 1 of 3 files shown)


external_test.go
```diff
@@ -1,7 +1,9 @@
 package quamina_test
```
timbray, line 1: I think this test would probably be better put in regexp_reader_test.go or maybe regexp_validity_test.go?  external_test.go is specifically aimed at testing the arguments of the top level Quamina data type.

external_test.go
```diff
@@ -50,3 +52,45 @@ func TestDifferentFlattener(t *testing.T) {
 		t.Error("missed!")
 	}
 }
+
+func TestRegexpCharacterClassAboveU10FFF(t *testing.T) {
+	const mathematicalBoldDigitZero = '\U0001D7CE'
+	if !unicode.Is(unicode.Nd, mathematicalBoldDigitZero) {
```
timbray: Why this check?

external_test.go
```diff
@@ -50,3 +52,45 @@ func TestDifferentFlattener(t *testing.T) {
...
+
+	q, err := quamina.New()
+	if err != nil {
+		t.Fatalf("New: %v", err)
+	}
+	patterns := []struct {
+		name    string
+		pattern string
+	}{
+		{name: "literal character class", pattern: fmt.Sprintf(`{"value": [{"regexp": "[%c]"}]}`, mathematicalBoldDigitZero)},
```
timbray: maybe put two characters inside the [] to wake up more of the regexp machinery? (not a blocker)

regexp_reader_test.go
```diff
@@ -88,6 +89,31 @@ func TestReadCCE1(t *testing.T) {
 	}
 }
 
+func TestIsCCcharBoundaries(t *testing.T) {
```
timbray, line 110: I like this!

timbray (commented):
This looks good - I have a couple of very minor comments.

external_test.go
Boulea7: Moved this coverage to `regexp_reader_test.go` as a focused reader test.

external_test.go
Boulea7: Removed this unrelated `unicode.Is` check and the property-matcher assertion.

external_test.go
Boulea7: Updated the character class to include two supplementary characters, `U+1D7CE` and `U+1D7CF`.

regexp_reader_test.go
Boulea7: Thanks — I kept the boundary assertions unchanged.

regexp_reader_test.go
```diff
@@ -88,6 +89,38 @@ func TestReadCCE1(t *testing.T) {
 	}
 }
 
+func TestReadCCESupplementaryCharacters(t *testing.T) {
+	_, err := readRegexp("[\U0001D7CE\U0001D7CF]")
+	if err != nil {
```
timbray, line 94: I think you used to have a test calling q.MatchesForEvent() to show that the regexp with supplementary characters actually worked?

If you could bring that back I'll approve and merge.

Boulea7: Restored an end-to-end `q.MatchesForEvent()` test in `regexp_reader_test.go`. It uses a character class with two supplementary characters and verifies that an event containing `U+1D7CE` produces the expected match; the unrelated category and property checks remain removed.

regexp_reader_test.go
```diff
@@ -89,10 +89,21 @@ func TestReadCCE1(t *testing.T) {
...
 
-func TestReadCCESupplementaryCharacters(t *testing.T) {
-	_, err := readRegexp("[\U0001D7CE\U0001D7CF]")
+func TestRegexpCharacterClassMatchesSupplementaryCharacter(t *testing.T) {
+	q, err := New()
 	if err != nil {
-		t.Fatalf("supplementary Unicode characters in character class: %v", err)
+		t.Fatalf("New: %v", err)
+	}
+	pattern := "{\"value\": [{\"regexp\": \"[\U0001D7CE\U0001D7CF]\"}]}"
```
timbray: One last comment and then we're good.  Per RFC8259, JSON literals are supposed to require \u rather than \U. I think it's actually a bug that Quamina accepts the \U but let's think about that separately.

Boulea7: Updated the test to use RFC 8259 `\u` surrogate pairs (`\uD835\uDFCE` and `\uD835\uDFCF`) in raw JSON strings. It still exercises the end-to-end `MatchesForEvent` path. I left the separate `\U` acceptance question out of this change.

timbray (commented):
LGTM. Will merge once the CI runs.

Curious: Did Claude find this one or did you run across it yourself?

timbray (approved):

Boulea7:
yeah, found it while chatting with Fable

timbray:
Thanks for the contribution!

Issue #554, opened 2026-06-22
Race in Pruner


timbray:
Pruner's `rebuildWhileLocked()` calls `memState.Iterate()` while its `MatchesForEvent()` calls `memState.Contains()`. `rebuildWhileLocked` defends against itself with a mutex, but not against `Contains`.

Also, a recent PR switched `memState` from a map to a slice to retain the order, in order to preserve `MatcherBuildMode` semantics, but `Containts` now traverses the whole slice; since `Contains` is called from `MatchesForEvent`, its performance is now `O(N)` where N is the number of patterns added.

timbray:
Fixed in #577

PR #577, opened 2026-09-01
fix: race condition in pruner

timbray:
addresses #554


timbray:
I'm ignoring the CI breakage for now. Codeql never fond anything useful anyhow.

Issue #560, opened 2026-06-22
Failure to treat -0 correctly


timbray:
pattern `{"x":  [0]}` matches `{"x":0}` but not `{"x":-0}`

Suggested fix: normalize `if f   == 0 { f = 0 }` at the top of numbitsFromFloat64

timbray:
fixed in #584

PR #584, opened 2026-09-07
fix: handle -0 correctly

timbray:
addresses #560


sayrer:
JSON supports it because IEEE 754 has it. 

https://en.wikipedia.org/wiki/Signed_zero

